PR #57, opened 2021-08-27
feat(http): add support for httpx

adriangb:
[HTTPX](https://www.python-httpx.org) is a requests-API compatible library providing high performance sync and async http clients.

This PR adds both sync and async clients backed by HTTPX.

Given that this client is brand new, maybe consider going with _just_ httpx so that you can offer support for both sync and async clients w/o the complexity or having 2 dependencies w/ vastly different interfaces?


adriangb:
Looking at this, I think (and please correct me) that the goal of having both requests and aiohttp is to provide sync and async clients. And the ability to give users the choice is to avoid having the extra dependency. And then in turn there is templating, abstract classes, etc. built around supporting this.

If this is the case, I think switching to _just_ httpx would simplify things a lot: since it's just 1 dependency, this package can just generate both sync & async clients at the same time: `from prisma import AsyncClient, Client`. There can be 2 separate interfaces for the sync and async clients, which simplifies the typing aspect. Templating can work pretty much as it does currently, except that both sync and async templates get generated.

What do you think?

adriangb:
I think this would also allow you to remove all of the code generation around QueryEngine, QueryBuilder and Client, leaving code generation only for the models themselves (which are agnostic to sync/async)

RobertCraigie:
> the goal of having both requests and aiohttp is to provide sync and async clients

You are correct.

> I think switching to just httpx would simplify things a lot

I do agree with that but mainly for the installation aspect, currently we require users to install prisma with an extra which I don't like as a naive first time user can't simply `pip install prisma-client`.

However I don't like switching to requiring a specific http library when we _could_ support multiple different http libraries without bloating users site packages. I do not know of a solution to this though.

> this package can just generate both sync & async clients at the same time

This is already theoretically possible as a user could install both requests and aiohttp at the same time. Generating both async and sync clients at the time is not a priority for me as the only use case I can see for this is typing client agnostic code e.g. a library that uses a user-generated prisma client, like [NextAuth](https://github.com/nextauthjs/next-auth) does but for python.

This is something I would like to support in the future though, feel free to create an issue for it.

> allow you to remove all of the code generation around QueryEngine, QueryBuilder and Client

- The QueryBuilder is a template as it requires usage of data that is only available after client generation.

- The QueryEngine could in theory be non-generated but this would probably lead to some code duplication which I do not like for user facing code.

- The Client must be generated as attributes are dynamic and rely on model names.

While we could refactor the QueryBuilder and the QueryEngine, making modules available pre-generation is not a priority for me.

RobertCraigie:
Also I'm not sure how I feel about adding support for another http library when the testing for `requests` is already sub-par, I will improve the testing for http libraries before merging this PR.

Thank you though, I really appreciate the effort you have put in to a client I released only a few days ago.

adriangb:
I did a bit of background reading, it seems like the Prisma Query Engine binary plans on moving away from HTTP interaction towards FFI. Which makes the choice of HTTP client a bit of moot point because it is only temporary. Does this sound correct?

RobertCraigie:
You are correct but there are no plans to move completely away from HTTP interaction as it is still needed for usage with the go client and there are no plans to use FFI with the go client anytime soon.

Interaction with the query engine binary using FFI is something that I will implement eventually but it is not a high priority for me as it is very complicated to get right.

> Which makes the choice of HTTP client a bit of moot point

I had not thought about it like that, I also think I'm overinflating the benefits of letting users choose the http library that is used internally.

When I have time I'll merge this and then make httpx the default and only http library we use as the only benefit of supporting multiple http libraries internally is reducing dependencies and as we'd have to install httpx anyway that would no longer apply.

RobertCraigie:
Will do, thanks.

From a quick glance through there doesn't seem to be too much to fix, the biggest thing I noticed was that you need to update the `http.py.jinja` template and `http.py` module to actually import your httpx code and update the generator config to support passing `httpx-async` and `httpx-sync`.

But given that I'm going to refactor a lot of the http stuff to make httpx the default anyway, I don't know if there's much point in cleaning it all up. It's up to you whether or not you want to get this PR fully working or just include the bare bones `_http_sync_http.py` and `_httpx_async_http.py` files.

src/prisma/_httpx_async_http.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class HTTP(AbstractHTTP[httpx.AsyncClient, httpx.Response]):
+    # pylint: disable=invalid-overridden-method,attribute-defined-outside-init
+
+    session: httpx.AsyncClient
+
+    async def download(self, url: str, dest: str) -> None:
+        resp: httpx.Response
+        async with self.session.stream("GET", url, timeout=None) as resp:
```
RobertCraigie, line 20:
```suggestion
        # pyright thinks that stream() doesn't return a context manager
        # due to how httpx handles compatability imports
        # https://github.com/encode/httpx/discussions/1829
        async with self.session.stream(
            'GET', url, timeout=None
        ) as resp:  # pyright: reportGeneralTypeIssues=false
```

src/prisma/_httpx_async_http.py
```diff
@@ -0,0 +1,76 @@
...
+            with open(dest, 'wb') as fd:
+                async for chunk in resp.aiter_bytes():
+                    fd.write(chunk)
+
+    async def request(self, method: Method, url: str, **kwargs: Any) -> 'Response':
+        return Response(await self.session.request(method, url, **kwargs))
```
RobertCraigie, line 26: Replace `data` with `content` due to deprecation warnings since httpx [v0.18.0](https://github.com/encode/httpx/blob/master/CHANGELOG.md#0180-27th-april-2021)
```suggestion
        content = kwargs.pop('data', None)
        if content is not None:
            kwargs['content'] = content

        return Response(await self.session.request(method, url, **kwargs))
```

src/prisma/_httpx_async_http.py
```diff
@@ -0,0 +1,76 @@
...
+
+class Response(AbstractResponse[httpx.Response]):
+    # pylint: disable=invalid-overridden-method
+
+    @property
+    def status(self) -> int:
+        return self.original.status_code
+
+    async def json(self, **kwargs: Any) -> Any:
+        return await self.original.json(**kwargs)
```
RobertCraigie, line 73: `httpx.Response.json()` isn't a coroutine
```suggestion
        return self.original.json(**kwargs)
```

src/prisma/_httpx_sync_http.py
```diff
@@ -0,0 +1,56 @@
...
+__all__ = ('HTTP', 'Response', 'client')
+
+
+class HTTP(AbstractHTTP[httpx.Client, httpx.Response]):
+    # pylint: disable=invalid-overridden-method,attribute-defined-outside-init
+
+    session: httpx.Client
+
+    def download(self, url: str, dest: str) -> None:
+        with self.session.stream("GET", url, timeout=None) as resp:
```
RobertCraigie, line 18: I prefer single quotes
```suggestion
        with self.session.stream('GET', url, timeout=None) as resp:
```

src/prisma/_httpx_sync_http.py
```diff
@@ -0,0 +1,56 @@
...
+            with open(dest, 'wb') as fd:
+                for chunk in resp.iter_bytes():
+                    fd.write(chunk)
+
+    def request(self, method: Method, url: str, **kwargs: Any) -> 'Response':
+        return Response(self.session.request(method, url, **kwargs))
```
RobertCraigie, line 24: Replace `data` with `content` due to deprecation warnings since httpx [v0.18.0](https://github.com/encode/httpx/blob/master/CHANGELOG.md#0180-27th-april-2021)
```suggestion
        content = kwargs.pop('data', None)
        if content is not None:
            kwargs['content'] = content

        return Response(self.session.request(method, url, **kwargs))
```

src/prisma/generator/templates/http.py.jinja
```diff
@@ -2,7 +2,11 @@
...
 {% if generator.config.http == 'aiohttp' %}
 from ._aiohttp_http import *
-{% else %}
+{% elif generator.config.http == 'requests' %}
 from ._requests_http import *
+{% elif generator.config.http == 'httpx-async' %}
+from ._httpx_async_http import *
+{# else: generator.config.http == 'httpx-sync' #}
+{% else %}
+from ._httpx_sync_http import *
```
RobertCraigie, line 11: Even though it's uninformative, raise an error if the target http library isn't handled here.
```suggestion
{% elif generator.config.http == 'httpx-sync' %}
from ._httpx_sync_http import *
{% else %}
{{ 0 / 0 }}
```

tests/integrations/httpx/schema.prisma
```diff
@@ -0,0 +1,16 @@
+datasource db {
+  provider = "sqlite"
+  url      = "file:dev.db"
+}
+
+generator client {
+  provider = "prisma-client-py"
+  http     = "httpx"
```
RobertCraigie, line 8:
```suggestion
  http     = "httpx-sync"
```

RobertCraigie (changes requested):
Thanks!

RobertCraigie:
@adriangb Would you be able to have a look at the changes I've requested? If not I'll happily commit them myself if that's okay with you?

adriangb:
Hey @RobertCraigie, sorry I haven't gotten around to it, got really busy all of a sudden. Please feel free to commit to my branch, or close this PR if you want to work on it in your own branch, anything goes, it's your repo 😄

RobertCraigie:
@adriangb No worries, thanks for letting me know about httpx!

PR #79, opened 2021-10-11
feat(client): validate query arguments at runtime

### LINKED ISSUE #60: Validate query arguments using pydantic
## Problem

Currently if any invalid arguments are passed then a very verbose and potentially confusing error message is raised, for example:

```prisma
model Post {
  id         String     @id @default(cuid())
  title      String
  published  Boolean
}
```

```py
await client.post.create({})
```

```
prisma.errors.MissingRequiredValueError: Failed to validate the query: `Unable to match input value to any allowed input type for the field. Parse errors: [Query parsing/validation error at `Mutation.createOnePost.data.PostCreateInput.published`: A value is required but not set., Query parsing/validation error at `Mutation.createOnePost.data.PostUncheckedCreateInput.published`: A value is required but not set.]` at `Mutation.createOnePost.data`
```

## Suggested solution

Use pydantic's `@validate_arguments` [decorator](https://pydantic-docs.helpmanual.io/usage/validation_decorator/).

The above example would then error with something like:

```
pydantic.error_wrappers.ValidationError: 2 validation errors for PostCreateInput
title
  field required (type=value_error.missing)
published
  field required (type=value_error.missing)
```

This feature should however have a schema option to disable validation as validation incurs a runtime performance cost and provides no benefits when static type checkers are used.

## Additional context

I suspect that the `@validate_arguments` decorator doesn't handle forward references properly so we'll probably have to do some horrible monkey patching to get this to work.

RobertCraigie:
closes #60

### DIFF
--- a/src/prisma/validator.py
+++ b/src/prisma/validator.py
@@ -4,6 +4,7 @@ from types import ModuleType
 from contextvars import ContextVar
 from functools import lru_cache, wraps
 from typing import Optional, Type, TypeVar, Iterator, Any, cast
+from typing_extensions import TypedDict
 
 from pydantic import BaseModel, Extra, create_model_from_typeddict
 from pydantic.decorator import validate_arguments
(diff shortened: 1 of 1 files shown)


RobertCraigie:
Blocked by https://github.com/samuelcolvin/pydantic/issues/3297

PR #98, opened 2021-11-05
chore: attempt to fix codecov early reports

RobertCraigie:
Currently, codecov is reporting it's status before all of our CI has passed, this is annoying as it means commits get flagged as failing when they are not.


RobertCraigie:
This did not work :(

RobertCraigie:
I do not know how to fix this :/

Will investigate again in the future or contact support

PR #102, opened 2021-11-06
chore: investigate mypy performance regression

RobertCraigie:


RobertCraigie:
Closing, identified the cause of the regression to be generic classes, https://github.com/python/mypy/issues/11484

PR #144, opened 2021-11-23
feat(ci): add CodeQL

RobertCraigie:
Don't know if I want to use this yet or not but GitHub recommended it so might as well give it a try.


RobertCraigie:
Decided against using CodeQL for now

PR #204, opened 2022-01-05
feat(http): increase default timeout to 30 seconds

### LINKED ISSUE #198: Increase default HTTP timeout
## Problem

Currently, `httpx.ReadTimeout` errors can be encountered easily (#197).

## Suggested solution

We should increase the default timeout, however I do not know by how much.

We should look into how Prisma handles this.

RobertCraigie:
closes #198


krrishdholakia:
Hey @RobertCraigie how can i override this? i'm trying to query a table with 1m+ rows and 30s causes timeouts

RobertCraigie:
@krrishdholakia here's an example https://github.com/RobertCraigie/prisma-client-py/pull/204/files

Issue #195, opened 2021-12-26
Support for ARM?


richin13:
## Bug description

Hello, I'm trying to run the `prisma` commands and to use the client on a raspberry pi that uses ARM64, I noticed the binaries that get downloaded are built for x86. Is there any way I can generate those binaries myself? Or download the appropriate version for my system?

## How to reproduce

Run an application on an ARM64 machine, running any command or attempting to use the client will result int:

```
$ python3 -m prisma db push
Traceback (most recent call last):
  File "/home/ubuntu/.asdf/installs/python/3.9.6/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/ubuntu/.asdf/installs/python/3.9.6/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/ubuntu/apps/bsc-tools/venv/lib/python3.9/site-packages/prisma/__main__.py", line 6, in <module>
    cli.main()
  File "/home/ubuntu/apps/bsc-tools/venv/lib/python3.9/site-packages/prisma/cli/cli.py", line 39, in main
    sys.exit(prisma.run(args[1:]))
  File "/home/ubuntu/apps/bsc-tools/venv/lib/python3.9/site-packages/prisma/cli/prisma.py", line 44, in run
    process = subprocess.run(
  File "/home/ubuntu/.asdf/installs/python/3.9.6/lib/python3.9/subprocess.py", line 505, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/home/ubuntu/.asdf/installs/python/3.9.6/lib/python3.9/subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/home/ubuntu/.asdf/installs/python/3.9.6/lib/python3.9/subprocess.py", line 1821, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
OSError: [Errno 8] Exec format error: '/tmp/prisma/binaries/engines/1c9fdaa9e2319b814822d6dbfd0a69e1fcc13a85/prisma-cli-linux'
```

Downloaded binary `file` result:

```
$ file /tmp/prisma/binaries/engines/1c9fdaa9e2319b814822d6dbfd0a69e1fcc13a85/prisma-cli-linux
[...]

PR #233, opened 2022-01-19
feat(binaries): improve platform and architecture support

HKGx:
## Change Summary

Resolves properly #195 
Pretty much just porting https://github.com/prisma/engines-wrapper to Python.
## Checklist

## Agreement


RobertCraigie:
Thank you! This is a much needed improvement.

Could you please add unit tests to ensure that all possible URLs that we generate will not result in a 404. You could base it off of this script (but should probably parallelise the requests with the `AsyncClient`):

```py
import httpx
from prisma.binaries import ENGINES

url = ENGINES[0].url
print(url)

client = httpx.Client()
response = client.head(url)
response.raise_for_status()

client.close()
```

HKGx:
> Could you please add unit tests to ensure that all possible URLs that we generate will not result in a 404. You could base it off of this script (but should probably parallelise the requests with the `AsyncClient`)

@RobertCraigie will do it later. Just pushed the basic implementation and now I'm going to sleep. It should be on par with the default TypeScript one. 
Tested it on Windows. Wanted to test on Debian but Prisma is not built for the newest OpenSSL 3.0 😩

RobertCraigie:
@HKGx No worries, thank you!

> Wanted to test on Debian but Prisma is not built for the newest OpenSSL 3.0 😩

I hadn't even realised there was an OpenSSL 3.0 to be honest, is it worth making an issue in https://github.com/prisma/prisma-engines?

RobertCraigie:
@HKGx I've pushed some minor changes and small fixes.

Some thoughts:

- I'm happy with the name change of `PRISMA_CLI_URL` to `PRISMA_CLI_MIRROR` and `PRISMA_ENGINES_MIRROR` but we need to document that somewhere (even if it wasn't documented before) as it is a breaking change.

Regarding code style:

- I prefer single quotes `'a'` over double quotes `"b"`
- I generally think if you need to create a method like: `engine_url_for` then that probably belongs within a class

Great work so far!

src/prisma/binaries/platform.py
```diff
@@ -1,75 +1,125 @@
...
+import sys
 from functools import lru_cache
-from typing import Tuple
-
-
-def name() -> str:
-    return _platform.system().lower()
+from typing import Optional
 
+import distro  # type: ignore # pyright: reportMissingTypeStubs=false
```
RobertCraigie, line 8:
```suggestion
import distro  # type: ignore[import]
```
This is a subtle change but means that both mypy and pyright will not report an error for missing type stubs but will still type check it. Previously pyright would disregard distros typings. You could either make this change wherever distro is imported or add an entry to `_compat.py`:

```py
import distro as distro  # type: ignore[import]
```
I haven't tested this but it should work.

PR #242, opened 2022-01-24
chore(cli/fetch): disable `--force` flag on windows

RobertCraigie:
## Change Summary

The `--force` flag is broken on windows.

Quoting HGKx:
> I've ran into a cheeky bug when using tempfiles. They all seem to implement O_EXCL flag which blocks opening them after they're created
and then we're using os.unlink to remove the binaries
but os.unlink tries to open the file?

This is not a high priority fix for me, however getting tests to pass on windows is.

## Checklist

## Agreement

### DIFF
--- a/src/prisma/cli/commands/fetch.py
+++ b/src/prisma/cli/commands/fetch.py
@@ -1,6 +1,6 @@
-import click
 import platform
 
+import click
 
 from ..utils import error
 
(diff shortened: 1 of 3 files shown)


RobertCraigie:
It turns out that it does actually work on windows, however it does not work in our unit tests ¯\_(ツ)_/¯

PR #255, opened 2022-01-30
feat(types): improve order by type safety

RobertCraigie:
## Change Summary

Improves the `order` type to signify that only one field can be present at any given time.

## Checklist

## Agreement

### DIFF
--- a/src/prisma/generator/schema.py
+++ b/src/prisma/generator/schema.py
@@ -121,6 +121,21 @@ class Model(BaseModel):
 
         return PrismaType.from_subtypes(subtypes, name=f'{model}WhereUniqueInput')
 
+    @cached_property
+    def order_by(self) -> PrismaType:
+        model = self.info.name
+        subtypes: List[PrismaType] = [
+            PrismaDict(
+                name=f'_{model}_{field.name}_OrderByInput',
+                total=True,
+                fields={
+                    field.name: 'SortOrder',
+                },
+            )
+            for field in self.info.scalar_fields
+        ]
+        return PrismaType.from_subtypes(subtypes, name=f'{model}OrderByInput')
+
 
 Schema.update_forward_refs()
 PrismaType.update_forward_refs()
--- a/src/prisma/generator/templates/types.py.jinja
+++ b/src/prisma/generator/templates/types.py.jinja
@@ -39,6 +39,35 @@ from .utils import _NoneType
     {% endif %}
 {% endmacro %}
 
+{% macro render_type(type) %}
+{# It is important that we render subtypes first so that references can be resolved correctly #}
+{% for subtype in type.subtypes %}
+{{ render_type(subtype) }}
+{% endfor %}
+{% if type.kind == 'alias' %}
+{{ type.name }} = {{ type.to }}
+{% elif type.kind == 'union' %}
+{{ type.name }} = Union[
+    {% for subtype in type.subtypes %}
+    '{{ subtype.name }}',
+    {% endfor %}
+]
+{% elif type.kind == 'typeddict' %}
+{# We use the old syntax for defined TypedDicts so that fields that shadow keywords #}
+{# can be defined, e.g. 'not' #}
+{{ type.name }} = TypedDict(
+    '{{ type.name }}',
+    {
+        {% for field, subtype in type.fields.items() %}
+        '{{ field }}': {{ type_as_string(subtype) }},
+        {% endfor %}
+    },
+    total={{ type.total }}
+)
+{% else %}
+{{ raise_err('Unhandled type kind: %s' % type.kind) }}
+{% endif %}
+{% endmacro %}
 
 SortMode = Literal['default', 'insensitive']
 SortOrder = Literal['asc', 'desc']
@@ -385,6 +414,7 @@ class _{{ type }}ListUpdatePush(TypedDict):
 {% endfor %}
 
 {% for model in dmmf.datamodel.models %}
+{% set model_schema = type_schema.get_model(model.name) %}
 # {{ model.name }} types
 
 class {{ model.name }}OptionalCreateInput(TypedDict, total=False):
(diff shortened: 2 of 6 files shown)


RobertCraigie:
If you need to order by multiple fields you can pass a list to `order`, for example:

```py
posts = await Post.find_many(
	order=[
        {'created_at': 'desc'},
        {'views': 'desc'},
    ],
)
```

danigosa:
Hi @RobertCraigie 

The solution is extremely inflexible and broke our code in development/testing branches, because it requires a `Literal` and something so simple as 

```python
def some_endpoint(sorty_by_field: str):
       posts = await Post.prisma().find_many(
	order=[
        {sort_by_field: 'desc'}
    ],
)
```

```
error: Argument "order" to "find_many" of "OrganizationActions" has incompatible type "Dict[str, Literal['asc', 'desc']]"; expected "Union[...]
```

Does not work. We tried to make Literals, Finals, but it fails on `mypy` and `pyright`. I do not know if there is a suitable solution for that, but writing the name of the order field as Literal in the code in advance is really making the use of `find_many` really hard with type checkers, the only solution for now for us is to bypass with `cast` or `type:ignore`.

danigosa:
Hi @RobertCraigie 

I can't find any `prisma.PostScalarFieldKeys`

image

Where this type is generated?

RobertCraigie:
@danigosa sorry I should've been more clear, that type is generated for every model in your schema file. So `PostScalarFieldKeys` is only generated if you have a model named `Post` in your schema file.

Are there other types like `PostScalarFieldKeys` generated?

danigosa:
> @danigosa sorry I should've been more clear, that type is generated for every model in your schema file. So `PostScalarFieldKeys` is only generated if you have a model named `Post` in your schema file.
> 
> Are there other types like `PostScalarFieldKeys` generated?

There is not, but it sounds great to have a Type that can handle this, can you post the code of `PostScalarFieldKeys` so I can work something out? There are not any generated for custom models, unfortunately

danigosa:
> @danigosa sorry I should've been more clear, that type is generated for every model in your schema file. So `PostScalarFieldKeys` is only generated if you have a model named `Post` in your schema file.
> 
> Are there other types like `PostScalarFieldKeys` generated?

Yes! I know documenting is hard, but this is really awesome and undocumented :laughing:

danigosa:
> > @danigosa sorry I should've been more clear, that type is generated for every model in your schema file. So `PostScalarFieldKeys` is only generated if you have a model named `Post` in your schema file.
> > Are there other types like `PostScalarFieldKeys` generated?
> 
> Yes! I know documenting is hard, but this is really awesome and undocumented laughing

Unfortunately with mypy it does not work:

```python
@router.get(
    "",
    response_model=Paginated[CenterWithoutRelations],
    response_class=ORJSONResponse,
)
async def list_centers(
    take: int = None,
    skip: int = None,
    sort_by: CenterScalarFieldKeys = "id",
    order: SortOrder = "asc",
    db: Prisma = Depends(db_from_request),
) -> Paginated[CenterWithoutRelations]:
    """LIST"""
    if sort_by not in CenterWithoutRelations.__fields__:
        return ORJSONResponse(
            status_code=400, content={"message": "Incorrect sort_by field"}
        )

    data = await db.center.find_many(
        take=take,
        skip=skip,
        where={"deleted": False},
        order={sort_by: order},
    )
    if skip is not None:
        count = await db.center.count()
    else:
        count = len(data)

    return {"data": data, "count": count}
```

```bash
➜  care-api git:(1446-care-api-dev-deploy) ✗ make stylecheck
Loading .env environment variables...
src/robbie/care_api/v1/centers.py:46: error: Argument "order" to "find_many" of "CenterActions" has incompatible type "Dict[Literal['id', 'org_id', 'name', 'created_at', 'modified_at', 'deleted'], Literal['asc', 'desc']]"; expected "Union[_Center_id_OrderByInput, _Center_org_id_OrderByInput, _Center_name_OrderByInput, _Center_created_at_OrderByInput, _Center_modified_at_OrderByInput, _Center_deleted_OrderByInput, List[Union[_Center_id_OrderByInput, _Center_org_id_OrderByInput, _Center_name_OrderByInput, _Center_created_at_OrderByInput, _Center_modified_at_OrderByInput, _Center_deleted_OrderByInput]], None]"
src/robbie/care_api/v1/centers.py:46: error: Expected TypedDict key to be string literal
```

The types code:

```python
CenterScalarFieldKeys = Literal[
    'id',
    'org_id',
    'name',
    'created_at',
    'modified_at',
    'deleted',
]
```

I think the problem is using `Union` of `TypedDict` with literal definition of key, may python typing does not support the goodies of TypeScript `type DerivedCenter = {[k: keyof Center]: string | number}` where you can define both types for keys and values... or I'm doing something wrong

danigosa:
Looking at https://peps.python.org/pep-0589/#rejected-alternatives I can see why is not working, `TypedDict` is not supposed to support that, any workaround is welcome!

danigosa:
Adding `ScalarFields` still have `mypy`  and `pyright` problems, the type in checkers after applying `ScalarFieldKeys` is `Dict[Literal['id', 'org_id', 'name', 'created_at', 'modified_at', 'deleted']` which is ok but does not work for `Union[TyopedDict,...]`:

```
src/robbie/care_api/v1/centers.py:46: error: Argument "order" to "find_many" of "CenterActions" has incompatible type "Dict[Literal['id', 'org_id', 'name', 'created_at', 'modified_at', 'deleted'], Literal['asc', 'desc']]"; 
expected "Union[_Center_id_OrderByInput, _Center_org_id_OrderByInput, _Center_name_OrderByInput, _Center_created_at_OrderByInput, _Center_modified_at_OrderByInput, _Center_deleted_OrderByInput, List[Union[_Center_id_OrderByInput, _Center_org_id_OrderByInput, _Center_name_OrderByInput, _Center_created_at_OrderByInput, _Center_modified_at_OrderByInput, _Center_deleted_OrderByInput]], None]"
src/robbie/care_api/v1/centers.py:46: error: Expected TypedDict key to be string literal
```
```    
"ORJSONResponse" is incompatible with "Paginated[CenterWithoutRelations]" (reportGeneralTypeIssues)
  /home/danigosa/rbcode/backend/care-api/src/robbie/care_api/v1/centers.py:46:15 - error: Argument of type "dict[str, str]" cannot be assigned to parameter "order" of type "CenterOrderByInput | List[CenterOrderByInput] | None" in function "find_many"
    Type "dict[str, str]" cannot be assigned to type "CenterOrderByInput | List[CenterOrderByInput] | None"
```

RobertCraigie:
> Yes! I know documenting is hard, but this is really awesome and undocumented 😆

Yeah the documentation needs some more love to be honest...

Any suggestions are more than welcome :)

> I think the problem is using Union of TypedDict with literal definition of key, may python typing does not support the goodies of TypeScript type DerivedCenter = {[k: keyof Center]: string | number} where you can define both types for keys and values... or I'm doing something wrong

I can replicate the errors you're seeing with both pyright and mypy. It saddens me to say this but I think this is just one of those cases where you'll have to ignore the error unfortunately. If you do this in multiple places I'd recommend extracting it to a function that returns the corresponding `PostOrderByInput` for your model so that you only have to ignore the error in that function.

@danigosa Thanks for looking into this and raising it, I was not aware that this did not work! This should probably also be raised on typing-sig / mypy / pyright to see if it would be feasible to support officially.

If I receive signal from other users that this change should be reverted, I'll look into it but for the time being I think this change provides a good amount of value as it signals to users directly writing the `order` query themselves that they can only specify one field at a time.

danigosa:
> > Yes! I know documenting is hard, but this is really awesome and undocumented laughing
> 
> Yeah the documentation needs some more love to be honest...
> 
> Any suggestions are more than welcome :)
> 
> > I think the problem is using Union of TypedDict with literal definition of key, may python typing does not support the goodies of TypeScript type DerivedCenter = {[k: keyof Center]: string | number} where you can define both types for keys and values... or I'm doing something wrong
> 
> I can replicate the errors you're seeing with both pyright and mypy. It saddens me to say this but I think this is just one of those cases where you'll have to ignore the error unfortunately. If you do this in multiple places I'd recommend extracting it to a function that returns the corresponding `PostOrderByInput` for your model so that you only have to ignore the error in that function.
> 
> @danigosa Thanks for looking into this and raising it, I was not aware that this did not work! This should probably also be raised on typing-sig / mypy / pyright to see if it would be feasible to support officially.
> 
> If I receive signal from other users that this change should be reverted, I'll look into it but for the time being I think this change provides a good amount of value as it signals to users directly writing the `order` query themselves that they can only specify one field at a time.

Yeah we are bypassing it with `cast(CenterOrderByInput, obj)` this way developers within the IDE can at least look into the target type to see what we are skipping by... 

After a talk with some team mates it's really a limitation on `typing`, although we do not find a TypeScript solution for it either, seems like the API from Query Engine is the one enforcing a bad typing by asking a **_list of dictionaries of just 1 pair per order field element_** which is itself a little weird, IMHO.

PR #267, opened 2022-02-03
refactor(client): rename `Client` to `Prisma`

### LINKED ISSUE #205: Rename `Client` to `Prisma`
## Problem

When using Prisma alongside other frameworks, it is not immediately clear what `Client` is a reference to - is it a HTTP Client? An external API Client?

## Suggested solution

We should rename the `Client` to `Prisma`, this makes it very clear. However we should also keep the `Client` name for backwards compatibility.

```py
class Prisma:
    ...

Client = Prisma
```

We should also update the documentation accordingly.

## Additional context

There is one downside to this, the python convention is to instantiate classes with the lowercase class name, e.g.

```py
prisma = Prisma()
```

As the package is also named `prisma` there could be name conflicts that arise from this.

kivo360:
## Change Summary

This was only a client name change. I made the changes inside of __init__.py. Instead of just referencing client I created an inherited client to update the representation. The class `Client` was changed to `Prisma` and the instance `client` was changed to `conn` within the docs. 

Closes #205.

## Checklist

## Agreement


RobertCraigie:
@kivo360 thank you!

Couple nitpicks: 

- I would prefer either a mix of using `db` and `prisma` or just `prisma` instead of `conn` as the client instance name in documentation / examples (I know I said there were downsides to this in the original issue but I do actually quite like using `prisma` as the name)
- the `prisma.http` module should not be commited unless you are actually changing code in the pre-generation state (this module is automatically replaced during generation time by the client)

What is the reasoning for inheriting from the Client class? I originally thought the best solution would be to rename it to Prisma and simply make an alias to the new class, e.g.

```py
class Prisma:
    ...

Client = Prisma
```

This means that anyone wanting to still use the old Client name can still do so with no changes as far as I am aware.

PR #289, opened 2022-02-16
chore(lint): move from black to blue for formatting

RobertCraigie:
## Change Summary

I like using single quotes for strings, blue supports automatically formatting strings to use single quotes while black does not. It also defaults to a different line width which is the most of the changes here.

## Checklist

## Agreement


q0w:
but black supports single-quote formatting (black -S) and line-length configurationю

RobertCraigie:
@q0w As far as I'm aware black doesn't support automatically formatting strings to use single quotes though. That's the value that blue provides for me. Blue is also built on top of black so switching isn't really a radical change either.

For example:

With the `-S` flag black will leave this code as is

```py
foo = 'bar'
bar = "baz"
````

Whereas blue will change it to

```py
foo = 'bar'
bar = 'baz'
```

PR #315, opened 2022-03-03
feat(package): add __slots__

### LINKED ISSUE #299: Add `__slots__` definitions
## Problem

We currently don't define the `__slots__` attribute for any of our classes. Adding it is a free performance improvement.

## Suggested solution

We should add `__slots__` definitions to as many classes as possible. https://github.com/ariebovenberg/slotscheck/ should then be included in tests to ensure we are defining the slots correctly.

Non-exhaustive list:
- `client.Prisma`
- `actions.*Actions`
- `http_abstract.HTTP`
- `http_abstract.Response`

I am unsure if we can/should add slots to model classes.

matyasrichter:
## Change Summary

Closes #299 
Adds `__slots__` to:
- `client.Prisma`
- `actions.*Actions`
- `http_abstract.HTTP` (+ sync and async subclasses)
- `http_abstract.Response` (+ sync and async subclasses)
- `QueryBuilder` + `Node` and its subclasses
- the `Base64` field type

I haven't added `__slots__` to models yet, but I would like to. The issue mentions that you're not sure about it, but I frankly can't see any problems it could cause. The `pydantic.BaseModel` superclass is slotted, and should someone need to subclass a model (though why would they?), slots on the model itself shouldn't cause any trouble.

## Checklist

## Agreement

### DIFF
--- a/src/prisma/_async_http.py
+++ b/src/prisma/_async_http.py
@@ -13,6 +13,8 @@ __all__ = ('HTTP', 'Response', 'client')
 class HTTP(AbstractHTTP[httpx.AsyncClient, httpx.Response]):
     session: httpx.AsyncClient
 
+    __slots__ = ()
+
     async def download(self, url: str, dest: str) -> None:
         async with self.session.stream('GET', url, timeout=None) as resp:
             resp.raise_for_status()
@@ -42,6 +44,8 @@ client: HTTP = HTTP()
 
 
 class Response(AbstractResponse[httpx.Response]):
+    __slots__ = ()
+
     @property
     def status(self) -> int:
         return self.original.status_code
--- a/src/prisma/_sync_http.py
+++ b/src/prisma/_sync_http.py
@@ -12,6 +12,8 @@ __all__ = ('HTTP', 'Response', 'client')
 class HTTP(AbstractHTTP[httpx.Client, httpx.Response]):
     session: httpx.Client
 
+    __slots__ = ()
+
     def download(self, url: str, dest: str) -> None:
         with self.session.stream('GET', url, timeout=None) as resp:
             resp.raise_for_status()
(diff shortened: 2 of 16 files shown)


RobertCraigie:
> I haven't added __slots__ to models yet, but I would like to. The issue mentions that you're not sure about it, but I frankly can't see any problems it could cause. 

My initial concerns where simply due to a lack of information, I did not know if pydantic would play well with `__slots__` or not.

> The pydantic.BaseModel superclass is slotted, and should someone need to subclass a model (though why would they?), slots on the model itself shouldn't cause any trouble.

People *might* want to add computed fields, e.g.

```py
class MyUser(User):
    @property
    def fullname(self) -> str:
        return f'{self.first_name} {self.surname}'
```

Although I do admit that the use cases for subclassing models are fairly niche.

I've tested it locally and directly instantiating and subclassing a model that has `__slots__` defined still works so feel free to go ahead and add `__slots__` to both models and partial models :)

Thanks for your contribution!

RobertCraigie (approved):

matyasrichter:
Hmm, looks like you were right to be concerned. When adding `__slots__` to the models template, I'm getting `AttributeError`s on model attribute access. I have to admit I don't understand what exactly is going on, but it seems like pydantic already works with __slots__ internally - to make their [private attributes](https://pydantic-docs.helpmanual.io/usage/models/#private-model-attributes) feature work. They construct `__slots__` and `__dict__` in `ModelMetaclass`.

Here's a minimal snippet that also raises `AttributeError` (python 3.10, pydantic 1.9.0):
```python
from pydantic import BaseModel

class A(BaseModel):
    name: str
    number: int

    __slots__ = ('name', 'number')


a = A(name="A", number=1)
a.dict()
a.name
```
```
Traceback (most recent call last):
  File "...", line 12, in <module>
    a.name
AttributeError: 'A' object has no attribute 'name'
```

I think we can simply merge this PR and not bother with slots on models for now, unless you've got some better insight on why exactly this isn't working.

RobertCraigie:
Thank you again! It is much appreciated 💜

PR #377, opened 2022-04-28
feat(client) add Decimal to is_number, Decimal group_by test, and pyp…

nilsso:
## Change Summary

Added `Decimal` to `is_number` to support `Decimal` fields in `group_by` aggregation.

## Checklist

## Agreement


pyproject.toml
```diff
@@ -1,3 +1,16 @@
...
+description = "Prisma Client Python is an auto-generated and fully type-safe database client"
+authors=["Robert Craigie <robert@craigie.dev>"]
+
+[tool.poetry.dependencies]
+python = "^3.10"
+
+[tool.poetry.dev-dependencies]
+pytest = "^7.1.2"
+syrupy = "^2.0.0"
+
```
RobertCraigie, line 13: While I would welcome a PR to switch to Poetry. I would rather not include anything poetry specific until we switch *everything* (or at least everything in the core package) to Poetry as it is something that I'll have to manually update and I would not be surprised if I forgot to do so.

tests/test_types/test_decimal.py
```diff
@@ -255,3 +257,34 @@ async def test_filtering_nulls(client: Prisma) -> None:
...
+    )
+
+    def assert_sum(i: int, v: str):
+        aggregate = aggregates[i]
+
+        assert '_sum' in aggregate
+        assert 'decimal_' in aggregate['_sum']
+        res = aggregate['_sum'][
+            'decimal_'
+        ]  # BUG: type should be str, not Decimal
```
RobertCraigie, line 286: This is a really unfortunate bug. As it would probably be quite difficult to fix, I'd be happy to merge this with this bug still present as long as the type definitions reflect the fact that a string is returned instead of a `Decimal` instance.

RobertCraigie (changes requested):
Thank you for working on this 💜

PR #378, opened 2022-04-28
feat(typing): change timeout args type to timedelta

### LINKED ISSUE #167: Time parameters should accept `datetime.timedelta` instead of `int`
## Problem

Currently timeout arguments are ambiguous as it is not clear what the value passed corresponds to. Is it seconds? Milliseconds?

## Suggested solution

We should refactor time parameter to accept a `datetime.timdelta` instance. This would also give the control back to the user, allowing them to ergonomically use the units of their desire.

We should also still accept integers but raise a deprecation warning and then remove them in the next release.

The-Daishogun:
change the type of timeout args  to timedelta
and raise a depreciation warning when provided type is an integer
fixes #167 
## Change Summary

change the type of timeout args in client generation and in `connect` method of client  to timedelta
and raise a depreciation warning when provided type is an integer

## Checklist

- ✅ Unit tests for the changes exist
- ✅ Tests pass without significant drop in coverage
- ✅ Documentation reflects changes where applicable
- ✅Test snapshots have been [updated](https://prisma-client-py.readthedocs.io/en/latest/contributing/contributing/#snapshot-tests) if applicable

## Agreement

### DIFF
--- a/src/prisma/generator/templates/client.py.jinja
+++ b/src/prisma/generator/templates/client.py.jinja
@@ -155,7 +155,7 @@ class Prisma:
         elif isinstance(connect_timeout, int):
             message = (
                 'Using integers for connect_timeout argument is deprecated'
-                ' and will be removed in the next release.\n'
+                ' and will be removed in the next major release.\n'
                 'You can disable this warning by passing a datetime.timedelta object instead of an integer. e.g.'
                 'prisma = Prisma(connect_timeout=datetime.timedelta(seconds=10))'
             )
@@ -222,7 +222,7 @@ class Prisma:
         elif isinstance(timeout, int):
             message = (
                 'Using integers for timeout argument is deprecated'
-                ' and will be removed in the next release.\n'
+                ' and will be removed in the next major release.\n'
                 'You can disable this warning by passing a datetime.timedelta object instead of an integer. e.g.'
                 'prisma.connect(timeout=datetime.timedelta(seconds=10))'
             )
--- a/src/prisma/generator/templates/client.py.jinja
+++ b/src/prisma/generator/templates/client.py.jinja
@@ -2,7 +2,7 @@
 {% from '_utils.py.jinja' import is_async, maybe_async_def, maybe_await, methods, operations, recursive_types, active_provider with context %}
 # -- template client.py.jinja --
 from types import TracebackType
-
+import warnings
 from . import types, models, errors, actions
 from .types import DatasourceOverride, HttpConfig
 from ._types import BaseModelT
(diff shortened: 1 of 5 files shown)


src/prisma/generator/templates/client.py.jinja
```diff
@@ -150,7 +150,19 @@ class Prisma:
         self._active_provider = '{{ active_provider }}'
         self._log_queries = log_queries
         self._datasource = datasource
-        self._connect_timeout = connect_timeout
+        if connect_timeout is None or isinstance(connect_timeout, UseClientDefault):
+            self._connect_timeout = datetime.timedelta(seconds=10)
+        elif isinstance(connect_timeout, int):
+            message = (
+                'Using integers for connect_timeout argument is deprecated'
+                ' and will be removed in the next release.\n'
```
RobertCraigie, line 158:
```suggestion
                ' and will be removed in the next major release.\n'
```

This would be a breaking change so I am going to hold off on enforcing this deprecation until the next major release.

src/prisma/generator/templates/client.py.jinja
```diff
@@ -199,14 +211,23 @@ class Prisma:
...
         """Connect to the Prisma query engine.
 
         It is required to call this before accessing data.
         """
         if isinstance(timeout, UseClientDefault):
             timeout = self._connect_timeout
+        elif isinstance(timeout, int):
+            message = (
+                'Using integers for timeout argument is deprecated'
+                ' and will be removed in the next release.\n'
```
RobertCraigie, line 225: See comment above.

tests/test_client.py
```diff
@@ -163,3 +163,18 @@ def test_old_client_alias() -> None:
     from prisma import Client, Prisma
 
     assert Client == Prisma
+
```
RobertCraigie, line 166: Could you also add tests to ensure that the correct number of seconds is passed to `engine.connect()`.

You could use the `patch_method` test function to do this, here's an example of how to use it: https://github.com/RobertCraigie/prisma-client-py/blob/main/tests/test_http.py#L67-L85

RobertCraigie (changes requested):
Thank you for working on this! 💜 

Just some minor nitpicks / changes.

The-Daishogun:
Thanks for the review.
I just have one question regarding the test case you recommended, pls check the corresponding conversation.

tests/test_client.py
The-Daishogun: aren't we testing the same thing in the following test?

https://github.com/RobertCraigie/prisma-client-py/blob/172b5f1bcbe91fda96027625b623f3ba41cd71e2/tests/test_client.py#L99-L113

The-Daishogun:
Sorry for the long delay. it's been a hectic couple of weeks.
i've made the requested changes, and would appreciate another review.

Issue #283, opened 2022-02-10
Support discarding action results


RobertCraigie:
## Problem

When doing batch inserts the resulting data is normally completely ignored. However we will still convert the data that Prisma gives us into Pydantic models, this incurs a significant runtime cost which could be avoided.

## Suggested solution

Add a new parameter to actions to signify that the resulting data can be discarded:

```py
@overload
async def create(
    ...
    ensure_result: Literal[True] = True,
) -> User:
    ...

@overload
async def create(
    ...
    ensure_result: Literal[False] = False,
) -> None:
    ...
```

## Alternatives

We should also limit the number of fields that are selected when this option is passed however that is out of scope for this issue.

## Additional context

See this discussion: https://github.com/RobertCraigie/prisma-client-py/issues/275#issuecomment-1034264260

leejayhsu:
@RobertCraigie I'm not seeing where pedantic objects are being created in batched queries. The only thing I can find is where the pydantic objects are created in non-batched `create`s ([here](https://github.com/RobertCraigie/prisma-client-py/blob/c988de2b1ee79124d6a65376492a4b9caaa0ea27/src/prisma/generator/templates/actions.py.jinja#L183))

is the above line where you'd like this new parameter to be, or am I looking in completely the wrong place? 🤔

RobertCraigie:
@leejayhsu I misspoke in my original comment. You are correct, the results of the query batcher are currently ignored. The only place pydantic objects are unconditionally created are in `actions.py.jinja` which you linked.

I also now think that `ensure_result` could be a confusing name, maybe `discard_result` is better?

leejayhsu:
@RobertCraigie thanks for confirming!

I like `discard_result` much more than `ensure_result` 😃 . I also like `return_result` (bonus points for the alliteration I guess)

PR #398, opened 2022-05-14
feat(client): support discarding query results

leejayhsu:
## Change Summary

This attempts to address https://github.com/RobertCraigie/prisma-client-py/issues/283 by introducing a new parameter `discard_result`.

When set to `True`, the pydantic objects will not be created from the results, and also `None` will be returned.
When not set or set to `False`, the original behavior is invoked (method will return pydantic objects).

## Checklist

## Agreement


leejayhsu:
@RobertCraigie [regarding this test](https://github.com/RobertCraigie/prisma-client-py/blob/acc5901711bed32b13a208386d5bd3d1367da12b/tests/test_batch.py#L244), should I add `discard_result` to `Batcher.create`'s signature, and ignore the arg?

RobertCraigie:
> @RobertCraigie [regarding this test](https://github.com/RobertCraigie/prisma-client-py/blob/acc5901711bed32b13a208386d5bd3d1367da12b/tests/test_batch.py#L244), should I add `discard_result` to `Batcher.create`'s signature, and ignore the arg?

Yes please. The current behaviour I'm going for is to maintain the same arguments for both batch queries and standard queries, even if it is redundant. I may change this in the future but for now I want to keep it like that.

src/prisma/generator/templates/actions.py.jinja
```diff
@@ -131,11 +132,55 @@ class {{ model.name }}Actions:
...
+        discard_result: Literal[True],
+    ) -> None:
+        ...
+
+    @overload
+    {{ maybe_async_def }}create(
+        self,
+        data: types.{{ model.name }}CreateInput,
+        *,
+        discard_result: Literal[False],
```
RobertCraigie, line 167: Looks like you ran into some issues with the ordering of arguments with defaults. If you mark all arguments after `data` as keyword only you should be able to remove some overloads and keep the `None` default for `include`.

RobertCraigie:
Thank you for working on this!

src/prisma/generator/templates/actions.py.jinja
leejayhsu: @RobertCraigie Sorry I'm not sure I follow.  Are you suggesting enforcing keywords by changing the `create` signature to:
```python
async def create(
        self,
        data: types.PostCreateInput,
        *,
        include: Optional[types.PostInclude] = None,
        discard_result: bool = False,
    ) -> Optional['models.Post']:
```

Or is the suggestion strictly about the overloads themselves?

Issue #22, opened 2021-06-19
Delegate binary downloading to prisma


RobertCraigie:
## Problem

With many possible configurations and settings, binary downloading is complicated, as such we should see if it is possible to delegate this task to the prisma as they already handle all of it for us.

This is an issue as we currently don't respect any binary targets options or http proxy options like prisma [does](https://www.prisma.io/docs/reference/api-reference/environment-variables-reference).

We would still have to download the CLI ourselves at the very least first.

RobertCraigie:
@janpio Thank you! I will definitely make use of that if I start working on this.

> prisma generate can also take care of downloading and supplying the Query Engine to a generator artifact

I did notice that while reading through some prisma code and added it but don't make use of it:
https://github.com/RobertCraigie/prisma-client-py/blob/33dc2c03e1357c0bf7d34d2a321815bc605aab4c/src/prisma/jsonrpc.py#L35

janpio:
My current understanding would be that it should pass a path to the engine, yes. But I am not aware of anyone using that, so might also be utterly broken. A quick read through the generator code on Prisma side might give clarity if this can even work.

RobertCraigie:
You are correct it does pass the engine path to the generator.

However even though the native binaries are successfully downloaded, the following error is thrown when running `prisma generate` with and without binary targets.

```
  prisma:loadEnv skipping package.json at /package.json +0ms
  prisma:tryLoadEnv Environment variables not found at null +0ms
  prisma:tryLoadEnv Environment variables not found at undefined +1ms
  prisma:tryLoadEnv No Environment variables loaded +0ms
  prisma:engines binaries to download query-engine, migration-engine, introspection-engine, prisma-fmt +0ms
Prisma schema loaded from tests/data/schema.prisma
  plusX Execution permissions of /private/var/folders/ql/0v8h20s972s6zz4t_3qc49bc0000gp/T/prisma-binaries/query-engine-darwin are fine +0ms
  prisma:getConfig Using Query Engine Binary at: /private/var/folders/ql/0v8h20s972s6zz4t_3qc49bc0000gp/T/prisma-binaries/query-engine-darwin +0ms
  plusX Execution permissions of /private/var/folders/ql/0v8h20s972s6zz4t_3qc49bc0000gp/T/prisma-binaries/query-engine-darwin are fine +505ms
[...]

PR #416, opened 2022-06-03
fix: use arm64 prisma binaries on arm64 machines

jordemort:
Fetch arm64 binaries when running on an arm64 machine.

This allows prisma-client-py to work on Linux and macOS machines with arm64 CPUs (it actually already worked on macOS because of Rosetta but this lets it use native binaries.)

Throw a RuntimeError if the machine is not arm64 or x86_64 to avoid downloading binaries with the wrong architecture.

Still WIP because Alpine seems broken, but I'm not 100% sure that Alpine was working to begin with; I'll poke at it more tomorrow.

## Change Summary

## Checklist

## Agreement

### DIFF
--- a/src/prisma/binaries/platform.py
+++ b/src/prisma/binaries/platform.py
@@ -1,13 +1,22 @@
-import re
-import sys
-import subprocess
 import platform as _platform
+import re
+import subprocess
+import sys
 from functools import lru_cache
 from typing import Tuple
 
 
 def name() -> str:
-    return _platform.system().lower()
+    system = _platform.system().lower()
+    machine = _platform.machine()
+
+    if machine in ("arm64", "aarch64"):
+        return f"{system}-arm64"
+
+    if machine != "x86_64":
+        raise RuntimeError(f"don't know how to fetch binaries for {machine}")
+
+    return system
 
 
 def check_for_extension(file: str) -> str:
@@ -50,7 +59,7 @@ def _get_linux_distro_details() -> Tuple[str, str]:
 @lru_cache(maxsize=None)
 def binary_platform() -> str:
     platform = name()
-    if platform != 'linux':
+    if platform not in ('linux', 'linux-arm64'):
         return platform
 
     distro = linux_distro()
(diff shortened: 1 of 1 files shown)


jordemort:
Ok, I don't think I've done any harm to alpine users here. I listed out the `prisma-photongo` bucket and tried all the likely binaries:

```
 <Key: prisma-photongo,prisma-cli-3.13.0-linux-arm64.gz>,
 <Key: prisma-photongo,prisma-cli-3.13.0-linux-x64.gz>,
 <Key: prisma-photongo,prisma-cli-3.13.0-linux.gz>,
```

None of those run on Alpine, even with `gcompat` installed.

Not sure how to best go about setting up a test for this. We could use QEMU to run an ARM64 container and run a test in there? It'd probably have to be a bag on the side of the current testing setup though.

iddan:
Seems like you need to run `blue` on `src/prisma/binaries/platform.py`

RobertCraigie:
Yes @iddan is correct. You can run `blue` by installing and running `pre-commit`, e.g.

```
pip install -U -r requirements/dev.txt
pre-commit install
pre-commit run --all-files
```

> None of those run on Alpine, even with gcompat installed.

What is the error message in this case?

> Not sure how to best go about setting up a test for this. We could use QEMU to run an ARM64 container and run a test in there? It'd probably have to be a bag on the side of the current testing setup though.

I'm not that fussed about having integration tests for this as I am looking into switching the implementation to directly use the Node CLI as that would allow us to remove all of the custom logic to download binaries. This shouldn't have any effect for users as we could automatically download node on machines that don't have it installed, see https://github.com/RobertCraigie/pyright-python for an example implementation of this behaviour and #22 for tracking.

jordemort:
It looks like https://github.com/RobertCraigie/prisma-client-py/pull/454 is going to make this completely obsolete, so I'm closing it.

RobertCraigie:
Hey @jordemort, I'm so sorry I didn't manage to get to your PR before it was superseded. I really appreciate the effort and I hope I can be more responsive if you decide to open another PR!

PR #419, opened 2022-06-04
fix(client): override the default SQLite relative file path

### LINKED ISSUE #400: Access to absolute path of schema file through client
## Problem

I want to locate the schema file my client is basing itself upon through the client, enabling me to find the absolute path of the schema.prisma file through python without much hassle. This would make debugging easier.

## Suggested solution

A function that returns the a path object pointing to the schema.prisma file the client is using.

## Alternatives

I'm not too sure, the generator could as well store the schema path in an environment variable while generation is occuring.

## Additional context

My actual problem isn't related to the client, but adding it would ease debugging. My libraries need to find a config file stored in the same directory as the schema, and prisma is my only dependency that generates a client based on a file, so my first thought was to see if prisma had such a function. (I can't use an absolute path because in production, it could be installed to either a global namespace or a virtual one, so i can't bank on knowing where my libraries will be)

Thanks for the good work Robert, I'm loving this client, it's my go to database solution for any project that needs one

### LINKED ISSUE #409: Prisma CLI and Client use different base paths for relative SQLite files
## Bug description

With this schema as `prisma/schema.prisma`:

```prisma
datasource db {
  provider = "sqlite"
  url      = "file:dev.db"
}

model User {
  id   String @id @default(cuid())
  name String
}
```

Running `prisma db push` creates a database file at `prisma/dev.db`.

Now running the client will error as the query engine uses the path relative to the working directory instead of the schema directory, e.g.

`main.py`
```python
import asyncio
from prisma import Prisma
from prisma.models import User

async def bar() -> None:
    p = Prisma(auto_register=True)
    await p.connect()
    await User.prisma().create(data={'name': 'Robert'})

asyncio.run(bar())
```

Will use `dev.db` instead of `prisma/dev.db`.

## Expected behavior

The CLI and the Client should use the same database path.

RobertCraigie:
## Change Summary

closes #400 
closes #409

This PR overrides the default SQLite URL to use a path relative to the Prisma Schema file instead of the working directory. This is a breaking change but is required in order to maintain consistency between the CLI and the Client.

This PR also exports a new constant, `SCHEMA_PATH` which points to the location of the schema file used to generate the client.

```py
from prisma import SCHEMA_PATH

print(SCHEMA_PATH)  # Path('/absolute/path/prisma/schema.prisma')
```

## Checklist

## Agreement

### DIFF
--- a/src/prisma/generator/models.py
+++ b/src/prisma/generator/models.py
@@ -304,7 +304,7 @@ class GenericData(GenericModel, Generic[ConfigT]):
     version: str
     generator: 'Generator[ConfigT]'
     dmmf: 'DMMF' = FieldInfo(alias='dmmf')
-    schema_path: str = FieldInfo(alias='schemaPath')
+    schema_path: Path = FieldInfo(alias='schemaPath')
     datasources: List['Datasource'] = FieldInfo(alias='datasources')
     other_generators: List['Generator[_ModelAllowAll]'] = FieldInfo(
         alias='otherGenerators'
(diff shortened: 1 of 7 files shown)


TheOnlyWayUp:
Awesome! The new constant will prove to be useful, and it's relieving that #409 has been fixed, ty Robert :)

PR #427, opened 2022-06-14
WIP: add support for transactions

### LINKED ISSUE #53: Add support for transactions
Prisma has preview support for [interactive transactions](https://www.prisma.io/docs/concepts/components/prisma-client/transactions#interactive-transactions).

The base work for this has already been done in the `wip/transactions` branch, some kinks need to be ironed out.

## Status

- [x] Model-based access support
- [ ] Protect against namespace collision

### Model-based access

We could refactor the internal strategy to register and retrieve clients to use contextvars. However we should only use contextvars for transactions as otherwise it would break certain async use cases like the `python -m asyncio` REPL.

RobertCraigie:
Opening prematurely to track status of tests / linters. When I have to occasionally update this branch I forget which things are "intended" to fail at the moment, while it is easy to just reset in git, it's nice to have it here too.

closes #53

TODO:
- [ ] call `rollback()` on error


listepo:
Any news?

RobertCraigie:
@listepo this is already supported, this PR is old and outdated. https://prisma-client-py.readthedocs.io/en/stable/reference/transactions/

PR #454, opened 2022-08-01
refactor(cli): remove pkg in favour of automatically downloading Node

### LINKED ISSUE #22: Delegate binary downloading to prisma

### LINKED ISSUE #236: Add support for ARM architecture for CLI binaries
## Problem

Currently the CLI binary targets are: `windows`, `linux` and `darwin`. We should also build for targets such as `node12-linux-arm64`

## Suggested solution

We do not currently have control over CLI binary building as this is handled and published by Prisma themselves. We will need to submit a PR in the Prisma Client Go repository.

The list of possible targets can be found here: https://github.com/vercel/pkg#targets.

## Additional context

https://github.com/RobertCraigie/prisma-client-py/issues/195#issuecomment-1001350680

This is also more relevant now that we are adding support for ARM engine binaries in #233

### LINKED ISSUE #370: Potentially unnecessary binary files on generate
## Bug description

In a basic Docker container:
```
FROM python:3.8
WORKDIR /app
COPY requirements.txt requirements.txt
RUN pip3 install -r requirements.txt
COPY schema.prisma schema.prisma
RUN prisma generate
COPY . .
CMD ["python3", "main.py"]
```

I was inspecting the Docker container file size using `dive`. Looking in the `/tmp` folder it looks like there may be some unused binaries. It appears that the binaries used are in `/tmp/prisma/binaries/engines/<version>` according to [constants.py](https://github.com/RobertCraigie/prisma-client-py/blob/main/src/prisma/binaries/constants.py#L41).

There is another folder `/tmp/prisma-binaries` with 114MB of binaries with the same name, but is missing the `prisma-cli-linux` but includes the others, e.g. `prisma-query-engine-debian-openssl-1.1.x`. If I remove this folder, the prisma query engine still appears to work.

It may be that the prisma generate and binary download process is downloading duplicate files.

## How to reproduce

```
docker build --platform linux/amd64 -t test-prisma
dive test-prisma
```

## Environment & setup

- OS: MacOS Docker building for linux/amd64
- Database: 
- Python version: 3.8
- Prisma version: prisma==0.6.4

### LINKED ISSUE #413: Move binaries outside of `/tmp` by default
## Problem

Currently, the downloaded binaries are placed in /tmp/ on Linux. However, /tmp/ directory is cleared on boot everytime.

## Suggested solution

Placing the binaries in /var/tmp/ on Linux. See [Temporary folder](https://en.wikipedia.org/wiki/Temporary_folder).

## Alternatives

Placing the binaries in site-packages/prisma/

### LINKED ISSUE #461: Can't generate prisma client with docker alpine
## Bug description
When running  prisma generate:

Downloading binaries  [####################################]  100%
Traceback (most recent call last):
  File "/usr/local/bin/prisma", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.10/site-packages/prisma/cli/cli.py", line 39, in main  
    sys.exit(prisma.run(args[1:]))
  File "/usr/local/lib/python3.10/site-packages/prisma/cli/prisma.py", line 76, in run
    process = subprocess.run(
  File "/usr/local/lib/python3.10/subprocess.py", line 501, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/local/lib/python3.10/subprocess.py", line 969, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/local/lib/python3.10/subprocess.py", line 1845, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/prisma/binaries/engines/efdf9b1183dddfd4258cd181a72125755215ab7b/prisma-cli-linux'

## How to reproduce

Run a python container with base image python:3.10.5-alpine3.16 and run prisma generate

## Hot to fix

Change to a debian based image example:
FROM python:3.10.5-slim-buster

### LINKED ISSUE #621: Error ENOENT on prisma generate
## Bug description

When I attempt to use prisma generate, I get an error message.

```
prisma generate

[DEBUG  ] prisma.binaries.binaries: query-engine cached at /tmp/prisma/binaries/engines/efdf9b1183dddfd4258cd181a72125755215ab7b/prisma-query-engine-debian-openssl-1.1.x
[DEBUG  ] prisma.binaries.binaries: migration-engine cached at /tmp/prisma/binaries/engines/efdf9b1183dddfd4258cd181a72125755215ab7b/prisma-migration-engine-debian-openssl-1.1.x
[DEBUG  ] prisma.binaries.binaries: introspection-engine cached at /tmp/prisma/binaries/engines/efdf9b1183dddfd4258cd181a72125755215ab7b/prisma-introspection-engine-debian-openssl-1.1.x
[DEBUG  ] prisma.binaries.binaries: prisma-fmt cached at /tmp/prisma/binaries/engines/efdf9b1183dddfd4258cd181a72125755215ab7b/prisma-prisma-fmt-debian-openssl-1.1.x
[DEBUG  ] prisma.binaries.binaries: prisma-cli-linux cached at /tmp/prisma/binaries/engines/efdf9b1183dddfd4258cd181a72125755215ab7b/prisma-cli-linux
[DEBUG  ] prisma.binaries.binaries: All binaries are cached
[DEBUG  ] prisma.cli.prisma: Using Prisma CLI at /tmp/prisma/binaries/engines/efdf9b1183dddfd4258cd181a72125755215ab7b/prisma-cli-linux
[DEBUG  ] prisma.cli.prisma: Running prisma command with args: ['generate']
Environment variables loaded from .env
Prisma schema loaded from schema.prisma
Error: Command failed with ENOENT: npm install -D prisma@3.13.0
spawn npm ENOENT

```

## How to reproduce

1. type `prisma generate` in command line

## Expected behavior

1. Colleagues appear to not have this same issue using the same commands and same version of prisma 

## Prisma information

```
// This is your Prisma schema file,
// learn more about it in the docs: https://pris.ly/d/prisma-schema

generator client {
  provider      = "prisma-client-js"
  binaryTargets = ["native", "linux-musl"]
}

datasource db {
  provider = "postgresql"
  url      = env("DATABASE_URL")
}

// unfortunately not usable yet https://graphql-nexus.github.io/nexus-prisma
// generator nexusPrisma {
//   provider = "nexus-prisma"
//   output   = "../generated/nexus-prisma"
// }

enum TrackType {
  TRUE_TRACKING
  BACK_TRACKING
  LIMIT_RAY_CASTING
  LIMIT_GCR
  STANDARD_GCR
  TRACE
}

enum SiteStatus {
  DRAFT
  SUBMITTED
  APPROVED
  REJECTED
  PARTS_ORDERED
  UNDER_CONSTRUCTION
  COMPLETED
}

enum Severity {
  PARTIAL_STOW
  FULL_STOW
}

enum AlarmTypes {
  WIND
  RAIN
  SNOW
  ICE
}

model Organization {
  id        String   @id @db.VarChar(20)
  createdAt DateTime @default(now())
  updatedAt DateTime @updatedAt
  sites               Site[]
  Zone                Zone[]
  Tracker             Tracker[]
  WindSensor          WindSensor[]
  Alarm               Alarm[]
  Device              Device[]
  PileLocation        PileLocation[]
  HorizonShadingPoint HorizonShadingPoint[]
  Schedule            Schedule[]
  ScheduleEntry       ScheduleEntry[]
}

model Site {
  id       String     @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
  name     String     @unique @db.VarChar(8)
  codename String     @unique @db.VarChar(50)
  status   SiteStatus @default(DRAFT)
  country  String?    @db.VarChar(2)
  state    String?    @db.VarChar(2)
  latitude  Float
  longitude Float
  elevation Float
  timezone  String @default("AmericaLos_Angeles")
  utcOffset Int    @default(-8)
  rotationLimit      Float?
  moduleLength       Float?
  minBearingSpacing  Float?
  moduleTorqeTubeGap Float?
  tracking           TrackType @default(TRUE_TRACKING)
  createdAt DateTime @default(now())
  updatedAt DateTime @updatedAt
  orgId        String       @default(dbgenerated("current_setting('user.org', true)::text"))
  organization Organization @relation(fields: [orgId], references: [id])
  zones      Zone[]
  trackers   Tracker[]
  devices    Device[]
  schedules  Schedule[]
  WindSensor WindSensor[]
  Alarm      Alarm[]
}

model Zone {
  id       String @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
  name     String
  number   Int
  position String
  // Network configuration
  staticIp      String?
  networkPrefix Int?
  gateway       String?
  // Other configuration
  scadaIp   String?
  zigBeePan String?
  ZoneController   Device @relation(fields: [zoneControllerId], references: [id])
  zoneControllerId String @db.Uuid
  Site   Site?   @relation(fields: [siteId], references: [id])
  siteId String? @db.Uuid
  orgId        String       @default(dbgenerated("current_setting('user.org', true)::text"))
  organization Organization @relation(fields: [orgId], references: [id])
  createdAt DateTime @default(now())
  updatedAt DateTime @updatedAt
  Tracker    Tracker[]
  WindSensor WindSensor[]
  @@unique([siteId, number])
}

model Tracker {
  id   String @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
  name String
  Site   Site?   @relation(fields: [siteId], references: [id])
  siteId String? @db.Uuid
  Zone   Zone?   @relation(fields: [zoneId], references: [id])
  zoneId String? @db.Uuid
  RowController   Device @relation(fields: [rowControllerId], references: [id])
  rowControllerId String @db.Uuid
  orgId        String       @default(dbgenerated("current_setting('user.org', true)::text"))
  organization Organization @relation(fields: [orgId], references: [id])
  createdAt    DateTime       @default(now())
  updatedAt    DateTime       @updatedAt
  PileLocation PileLocation[]
  @@unique([siteId, name])
  @@unique([siteId, zoneId, name])
}

model WindSensor {
  id       String @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
  name     String
  Site     Site   @relation(fields: [siteId], references: [id])
  siteId   String @db.Uuid
  Zone     Zone   @relation(fields: [zoneId], references: [id])
  zoneId   String @db.Uuid
  Device   Device @relation(fields: [deviceId], references: [id])
  deviceId String @db.Uuid
  orgId        String       @default(dbgenerated("current_setting('user.org', true)::text"))
  organization Organization @relation(fields: [orgId], references: [id])
  createdAt DateTime @default(now())
  updatedAt DateTime @updatedAt
}

model Alarm {
  id   String     @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
  type AlarmTypes @default(WIND)
  name String
  triggerSpeed Float
  triggerTime  Int
  clearSpeed   Float
  clearTime    Int
  severity     Severity
  Site   Site   @relation(fields: [siteId], references: [id])
  siteId String @db.Uuid
  orgId        String       @default(dbgenerated("current_setting('user.org', true)::text"))
  organization Organization @relation(fields: [orgId], references: [id])
  createdAt DateTime @default(now())
  updatedAt DateTime @updatedAt
}

model Device {
  id String @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
  // Hardware identification
  ethernetMac  String?
  serialNumber String?
  zigBeeMac    String?
  Site   Site   @relation(fields: [siteId], references: [id])
  siteId String @db.Uuid
  orgId        String       @default(dbgenerated("current_setting('user.org', true)::text"))
  organization Organization @relation(fields: [orgId], references: [id])
  createdAt DateTime @default(now())
  updatedAt DateTime @updatedAt
  Tracker    Tracker[]
  Zone       Zone[]
  WindSensor WindSensor[]
}

model PileLocation {
  id    String @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
  label String
  x Float
  y Float
  z Float
  moduleNumber Int @default(0)
  inverter     Int @default(0)
  Tracker   Tracker @relation(fields: [trackerId], references: [id])
  trackerId String  @db.Uuid
  orgId        String       @default(dbgenerated("current_setting('user.org', true)::text"))
  organization Organization @relation(fields: [orgId], references: [id])
  createdAt DateTime @default(now())
  updatedAt DateTime @updatedAt
}

model HorizonShadingPoint {
  id String @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
  azimuth   Float
  elevation Float
  zenith    Float
  x         Float?
  y         Float?
  z         Float?
  Schedule   Schedule? @relation(fields: [scheduleId], references: [id])
  scheduleId String?   @db.Uuid
  orgId        String       @default(dbgenerated("current_setting('user.org', true)::text"))
  organization Organization @relation(fields: [orgId], references: [id])
  createdAt DateTime @default(now())
  updatedAt DateTime @updatedAt
}

model Schedule {
  id        String    @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
  type      String?
  version   String
  trackType TrackType
  Site   Site   @relation(fields: [siteId], references: [id])
  siteId String @db.Uuid
  orgId        String       @default(dbgenerated("current_setting('user.org', true)::text"))
  organization Organization @relation(fields: [orgId], references: [id])
  createdAt DateTime @default(now())
  updatedAt DateTime @updatedAt
  horizon  HorizonShadingPoint[]
  times    ScheduleEntry[]
  trackers String[]
}

model ScheduleEntry {
  id String @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
  Schedule   Schedule @relation(fields: [scheduleId], references: [id])
  scheduleId String   @db.Uuid
  orgId        String       @default(dbgenerated("current_setting('user.org', true)::text"))
  organization Organization @relation(fields: [orgId], references: [id])
  day    Int
  month  Int
  hour   Int
  minute Int
  values Int[]
}
```

## Environment & setup

- OS: Ubuntu 20.04
- Database: Postgresql
- Python version: 3.8.10
- Prisma version:

```
prisma                : 3.13.0
@prisma/client        : Not found
Current platform      : debian-openssl-1.1.x
Query Engine (Binary) : query-engine efdf9b1183dddfd4258cd181a72125755215ab7b (at ../../../../../tmp/prisma/binaries/engines/efdf9b1183dddfd4258cd181a72125755215ab7b/prisma-query-engine-debian-openssl-1.1.x, resolved by PRISMA_QUERY_ENGINE_BINARY)
Migration Engine      : migration-engine-cli efdf9b1183dddfd4258cd181a72125755215ab7b (at ../../../../../tmp/prisma/binaries/engines/efdf9b1183dddfd4258cd181a72125755215ab7b/prisma-migration-engine-debian-openssl-1.1.x, resolved by PRISMA_MIGRATION_ENGINE_BINARY)
Introspection Engine  : introspection-core efdf9b1183dddfd4258cd181a72125755215ab7b (at ../../../../../tmp/prisma/binaries/engines/efdf9b1183dddfd4258cd181a72125755215ab7b/prisma-introspection-engine-debian-openssl-1.1.x, resolved by PRISMA_INTROSPECTION_ENGINE_BINARY)
Format Binary         : prisma-fmt efdf9b1183dddfd4258cd181a72125755215ab7b (at ../../../../../tmp/prisma/binaries/engines/efdf9b1183dddfd4258cd181a72125755215ab7b/prisma-prisma-fmt-debian-openssl-1.1.x, resolved by PRISMA_FMT_BINARY)
Default Engines Hash  : efdf9b1183dddfd4258cd181a72125755215ab7b
Studio                : 0.459.0

```

RobertCraigie:
## Change Summary

This PR completely refactors how the Prisma CLI is downloaded / installed / called. We now download Node itself at runtime and use that to install the Prisma CLI and then run it directly with Node as well.

This has some significant advantages:

- We are now no longer tied to releases of the packaged CLI
	- These were being released by Luca, who created the Go Client and is no longer at Prisma, on his own free time.
	- Only major versions were released which means the CLI couldn't be easily tested with the latest changes on the https://github.com/prisma/prisma repository
- Prisma Studio can now be launched from the CLI
- The TypeScript Client can now be generated from our CLI wrapper
- We now longer have to manually download the engine binaries ourselves, that's handled transparently for us
- A packaged version of Node no longer has to be installed for each new Prisma version
- We now have support for ARM

However, this does not come without some concerns:

- Size increase
	- We use https://github.com/ekalinin/nodeenv to download Node at runtime if it isn't already installed. This downloads and creates extra files that are not strictly necessary for our use case. This results in an increase from ~140MB -> ~300MB. - 
	- However this size increase can be reduced by installing [nodejs-bin](https://pypi.org/project/nodejs-bin/) which you can do by providing the `node` extra, e.g. `pip install prisma[node]`. This brings the total download size to be very similar to the packaged CLI.
	- This concern also doesn't apply in cases where Node is already present. It actually will greatly improve the experience in this case.

## How does it work?

We now resolve a Node binary using this flow:

- Check if [nodejs-bin](https://pypi.org/project/nodejs-bin/) is installed
- Check if Node is installed globally
- Downloads Node using https://github.com/ekalinin/nodeenv to a configurable location which defaults to `~/.cache/prisma-nodeenv`

The first two steps in this flow can be skipped if you so desire through your `pyproject.toml` file or using environment variables. For example:

```toml
[tool.prisma]
# skip global node check
use_global_node = false

# skip nodejs-bin check
use_nodejs_bin = false

# change nodeenv installation directory
nodeenv_cache_dir = '~/.foo/nodeenv'
```

Or using the environment variables, `PRISMA_USE_GLOBAL_NODE`, `PRISMA_USE_NODEJS_BIN` and `PRISMA_NODEENV_CACHE_DIR` respectively. 

The Prisma CLI is then installed directly from [npm](https://www.npmjs.com/package/prisma) and ran directly using the resolved Node binary.

closes #22
closes #236
closes #370 
closes #413
closes #461
closes #621

## Checklist

TODO:
- [x] Write node tests
- [x] Ensure we don't use an out of date global node installation
- [x] Fix windows
- [x] Documentation
	- [x] Document config
- [x] Fix TODOs in `_node.py`
- [x] Fix incorrect caching location in tests
- [x] Fix `prisma_version` not reading from environment variables
- [x] Audit subprocess calls
- [x] Test refactored fetch command
- [x] Check `binaryTargets` warning.

## Agreement


medz:
Seems like a good idea. The biggest advantage I think is to save the development of the CLI part. I had this in mind when developing [Prisma ORM for Dart](https://github.com/odroe/prisma-dart). However, adding the Node environment in the current development environment seems to be redundant and disgusting.

medz:
😊Wait for you to add this function experimentally first, if the user feedback is not disgusting to download the Node environment, I think Prisma client development in other languages can be used for reference.

RobertCraigie:
> However, adding the Node environment in the current development environment seems to be redundant and disgusting.

@medz Are you referring to the fact that Node will be installed into the user's environment? If so, that is not the case, Node will be downloaded to a temporary / cache directory that is not included in the user's `PATH` or environment setup in any way.

> [Prisma ORM for Dart](https://github.com/odroe/prisma-dart)

Wow! I'm excited to hear that Prisma has been ported to yet another language! Well done and I hope it goes well :)

Let me know if you run into any issues with your integration, I'm more than happy to help!

jacobdr:
I think this makes a lot of sense and KISS in terms of keeping track with the upstream project more easily which is pretty NodeJS-centric.

Question: why not make the `prisma[node]` the default to quell the size increase concerns?

jacobdr:
Something somewhere is choking on the installation via `.[node]` expecting `npm` and `node` on the PATH. Need to track that down a bit more

```bash
npm@8.12.1 /home/prisma/.local/lib/python3.10/site-packages/nodejs/lib/node_modules/npm
prisma@c7ca946a4a90:~/.local/lib/python3.10/site-packages/nodejs$ /home/prisma/.local/lib/python3.10/site-packages/nodejs/bin/node /home/prisma/.local/lib/python3.10/site-packages/nodejs/lib/node_modules/npm/bin/npm-cli.js install prisma@3.13.0
npm notice
npm notice New minor version of npm available! 8.12.1 -> 8.19.2
npm notice Changelog: https://github.com/npm/cli/releases/tag/v8.19.2
npm notice Run npm install -g npm@8.19.2 to update!
npm notice
npm ERR! code 127
npm ERR! path /home/prisma/.local/lib/python3.10/site-packages/nodejs/node_modules/prisma
npm ERR! command failed
npm ERR! command sh -c node scripts/preinstall-entry.js
npm ERR! sh: 1: node: not found

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/prisma/.npm/_logs/2022-10-16T20_57_02_337Z-debug-0.log
```

jacobdr:
Update to the above -- its coming from the `@prisma/engines` somewhere, but still need to track that down even more:

```bash
~/.local/lib/python3.10/site-packages/nodejs$ /home/prisma/.local/lib/python3.10/site-packages/nodejs/bin/node /home/prisma/.local/lib/python3.10/site-packages/nodejs/lib/node_modules/npm/bin/npm-cli.js install @prisma/engines@4.4.0

npm ERR! code 127
npm ERR! path /home/prisma/.local/lib/python3.10/site-packages/nodejs/node_modules/@prisma/engines
npm ERR! command failed
npm ERR! command sh -c node scripts/postinstall.js
npm ERR! sh: 1: node: not found

```

jacobdr:
Actually, I think its really coming from the @prisma/client dependency:

```bash
DEBUG=* python -m nodejs.npm install --foreground-scripts @prisma/client

```

requirements/node.txt
```diff
@@ -0,0 +1 @@
+nodejs-bin
```
jacobdr, line 1: If we set this to `nodejs-bin[cmd]` that should put `npm` and `node` on the PATH but we'll need to call out a note re: collisions.

Alternatively we can probably intercept the PATH at some point during the installation sequence and CLI invocation shims in this PR and point to the version provided by `nodejs-bin` within our site-packages

See https://pypi.org/project/nodejs-bin/

jacobdr (commented):
I think we need to add `nodeenv` as a regular dependency

RobertCraigie:
> Question: why not make the prisma[node] the default to quell the size increase concerns?

The only reason I wouldn't want to do that is because making it optional means that on machines that already have node we will be redundantly downloading Node again when we *could* just use the already installed Node.

> Something somewhere is choking on the installation via .[node] expecting npm and node on the PATH. Need to track that down a bit more

Oh this is an interesting bug, thank you for finding it before this was merged!

jacobdr:
Fair enough re: your point above **but** if someone does not have NodeJS a first-time user is going to have a worse experience with a pip install. Also, anyone who is currently using this in CI is likely to experience this as a breaking change unless they updated their own dependency on us to specify the `[node]` extra. 

If anything I think it might be safer to download a controlled NodeJS version ourselves and invert the logic -- NOT installing node if that is specified as an extra. Not sure if that is even really possible in the Python packaging ecosystem but I guess we would look into it.

Granted, everything is pre 1.0 right now but given the popularity you have already I think we would not want to squander it with a breaking change.

src/prisma/generator/models.py
```diff
@@ -347,6 +350,17 @@ def validate_version(cls, values: Dict[Any, Any]) -> Dict[Any, Any]:
         return values
 
 
+class BinaryPaths(BaseModel):
+    query_engine: Dict[str, str] = FieldInfo(
```
jacobdr, line 354: Maybe a comment here about what this mapping is would be useful? I assume its something that platform to filename of the engine? An alternative to not using code would be a semantic type alias that resolves to str but is named something meaningful like `PLATFORM`. Or we can use enum as key?

src/prisma/cli/prisma.py
```diff
@@ -34,60 +28,60 @@ def run(
...
-                ' ' * 3,
-            ),
-            nl=False,
-        )
-        click.echo(click.style('npx prisma studio', bold=True))
-        return 1
-
-    process = subprocess.run(
-        [str(path.absolute()), *args],
+    # TODO: ensure graceful termination
```
jacobdr, line 32: One idea: You could maybe just make the ABC implementation responsible for the actual calling and have the implementation `run` methods just pass through their arguments and all the other foo from the signature. So that subprocess is called in just one place -- as a private method call on `Strategy`. That would force all exit code checking into one code path. Otherwise each strategy needs to concern itself not with building args but also with error checking

src/prisma/engine/utils.py
```diff
@@ -29,20 +32,37 @@
 }
 
 
-def ensure() -> Path:
+def query_engine_name() -> str:
+    return f'prisma-query-engine-{platform.check_for_extension(platform.binary_platform())}'
+
+
+# TODO: detect this smarter, search for the current platform name?
+def _resolve_from_binary_paths(binary_paths: dict[str, str]) -> Path | None:
```
jacobdr, line 40: I think the type of the `dict` here should be `Dict` (unless type checkers now allow this?)

src/prisma/_config.py
```diff
@@ -46,6 +41,27 @@ class DefaultConfig(BaseSettings):
...
+    # Extra arguments to pass to nodeenv, arguments are passed after the path, e.g. python -m nodeenv <path> <extra args>
+    nodeenv_extra_args: List[str] = Field(
+        env='PRISMA_NODEENV_EXTRA_ARGS',
+        default_factory=list,
+    )
+
+    # Where to download nodeenv to, defaults to ~/.cache/prisma-nodeenv
+    nodeenv_cache_dir: Path = Field(
+        env='PRISMA_NODEENV_CACHE_DIR',
+        default_factory=lambda: Path.home() / '.cache' / 'prisma-nodeenv',
```
jacobdr, line 62: If we are going to start owning our own cache (unless this is created by nodeenv), I think we should have the top level be our own package name similar to what prisma does with its `.prisma` directory. Same comment applies to the `'.cache' / 'prisma-binaries'` below

jacobdr:
@RobertCraigie I know you still have some TOODs but thought I would leave an early review. Let me know when you think this is ready for me to check out and run locally. We're going to have to test the shit out of this PR

src/prisma/engine/utils.py
RobertCraigie: In Python 3.9+ or in Python 3.7+ with `from __future__ import annotations` you can use the actual runtime types in type hints like this.

RobertCraigie:
> Also, anyone who is currently using this in CI is likely to experience this as a breaking change unless they updated their own dependency on us to specify the [node] extra.

This shouldn't be a breaking change for anyone. I think you might've thought it would be as I actually forgot to include `nodeenv` as a base dependency in this PR which I hadn't realised until now.

> If anything I think it might be safer to download a controlled NodeJS version ourselves and invert the logic -- NOT installing node if that is specified as an extra. Not sure if that is even really possible in the Python packaging ecosystem but I guess we would look into it.

Yeah I gavelled about this while working on this PR initially - it is not possible to have "default" extras which means we wouldn't be able to provide an option to *not* install Node even if it is already installed.

----

I think a good solution to this problem is to document the `node` extra as much as possible, e.g. wherever we document how to install the package, it also includes `[node]`, that way it is very visible and hopefully most first time users would install it this way and anyone who cares about not installing Node to `site-packages` can just omit the `node` extra.

src/prisma/cli/prisma.py
RobertCraigie: Good idea - used this implementation strategy in 694dffbe343fd8930afd8423067248483398f71a.

jacobdr:
Interestingly, we actually don't need the node dependency at runtime, right? Not to throw more confusion into the mix, but at "production" runtime (not dev) I believe we only need the generated Python code and the WASM engine binary. Just thinking through attack surface area, dragging in node+npm is pretty crappy (totally disregarding the size concern).

To be honest I don't quite have a solution yet to that as I need to mull it over more, but wanted to call it out (and please do correct me if I am wrong)

src/prisma/_config.py
jacobdr: Or re-use the same caching directory used by the upstream project to reduce the overhead of download for devs operating across both ecosystems. Not sure how common that is, but I fall into the category

src/prisma/cli/_node.py
```diff
@@ -0,0 +1,320 @@
...
+from pydantic.typing import get_args
+
+from .. import config
+from .._proxy import LazyProxy
+from ..binaries import platform
+from ..errors import PrismaError
+
+if TYPE_CHECKING:
+    # TODO
+    import nodejs  # type: ignore
```
jacobdr, line 22: I have a WIP to add typings to this project for us: https://github.com/jacobdr/nodejs-pypi/pull/4 but haven't heard from the maintainer yet

src/prisma/cli/_node.py
```diff
@@ -0,0 +1,320 @@
...
+    def __run__(
+        self,
+        *args: str,
+        check: bool = False,
+        cwd: Path | None = None,
+        stdout: File | None = None,
+        stderr: File | None = None,
+        env: Mapping[str, str] | None = None,
+    ) -> subprocess.CompletedProcess[bytes]:
+        print('--- debug ---')
```
jacobdr, line 125: Reminder to remove this before merge (and below)

src/prisma/cli/_node.py
```diff
@@ -0,0 +1,320 @@
...
+    if env is None:
+        env = dict(os.environ)
+
+    assert target_bin.exists(), 'Target `bin` directory does not exist'
+
+    path = env.get('PATH', '') or os.environ.get('PATH', '')
+    if path:
+        # handle the case where the PATH already ends with the `:` separator (this probably shouldn't happen)
+        if path.endswith(':'):
+            path = f'{path}{target_bin.absolute()}'
```
jacobdr, line 282: Just wondering if we should prefix our node version vs adding it to the end... That would give more control over our code to a known `node` or `npm` version that we have brought in ourselves via the `prisma-py[node]` install method. Probably a lot less room for weird bugs that way. At least on my system, this would mean even though I installed the `prisma[node]` way, it would still resolve to whatever my global `node` was which would sit earlier on the PATH

src/prisma/cli/_node.py
```diff
@@ -0,0 +1,320 @@
...
+Node = Union[NodeJSPythonStrategy, NodeBinaryStrategy]
+
+
+def resolve(target: Target) -> Node:
+    if target not in {'node', 'npm'}:
+        raise UnknownTargetError(target=target)
+
+    if config.use_nodejs_bin:
+        log.debug('Checking if nodejs-bin is installed')
+        if nodejs is not None:
```
jacobdr, line 258: Do we need an error conditional here in an `else` statement? Or is that checked at a higher level?

src/prisma/cli/_node.py
RobertCraigie: Thanks! That PR should only effect internal typing right?

The package was previously missing a `py.typed` file which is the original reason for these `# type: ignore`s. I [created a PR](https://github.com/samwillis/nodejs-pypi/pull/8) to add it which was merged and released so these may be unnecessary now.

jacobdr: Did we turn on the mypy setting to turn unused ignores into errors? That would answer that question. And with re: to external types, yeah its only a partial solution so far, but eventually I'll add external types to that project

src/prisma/cli/_node.py
RobertCraigie: No, with the current structure `use_nodejs_bin` being set to `True` is just the default, it's intended to indicate whether or not we should attempt to use it, if it is available. It's not meant to indicate that it is the only option.

Should this config option be renamed to make this more clear?

src/prisma/cli/_node.py
```diff
@@ -0,0 +1,320 @@
...
+
+        The interface for this function is very similar to `subprocess.run()`.
+        """
+        return self.__run__(
+            *args,
+            check=check,
+            cwd=cwd,
+            stdout=stdout,
+            stderr=stderr,
+            env=_update_path_env(
```
jacobdr, line 67: Can you move this up and print it at the debug level? Might help us from a logging standpoint understand the CI failure -- https://github.com/RobertCraigie/prisma-client-py/actions/runs/3400416165/jobs/5654829166#step:3:1336

src/prisma/cli/_node.py
```diff
@@ -0,0 +1,320 @@
...
+
+                print(p)
+        print('--- end ---')
+        print('-- list dir --')
+        for p in self.path.parent.iterdir():
+            print(p)
+        print('-- end --')
+
+        path = str(self.path.absolute())
+        log.debug('Executing binary at %s with args: %s', path, args)
```
jacobdr, line 139: This is more an aside... Why this format over f-strings?

src/prisma/cli/_node.py
```diff
@@ -0,0 +1,320 @@
...
+        func = None
+        if self.target == 'node':
+            func = nodejs.node.run
+        elif self.target == 'npm':
+            func = nodejs.npm.run
+        else:
+            raise UnknownTargetError(target=self.target)
+
+        # TODO
+        return func(  #  type: ignore
```
jacobdr, line 235: We have logging of the invocation in the `NodeBinaryStrategy` -- I reckon we should add that debug output here as well before we make the function call to ease debugging

src/prisma/cli/_node.py
RobertCraigie: I think a better solution would be to add debug logging directly in the `_update_path_env` function - thanks for pointing out that this needs better logging :)

Also I think the issue is that I had used `:` as the path separator when windows actually uses `;`, hopefully that's fixed now.

src/prisma/cli/_node.py
RobertCraigie: It's so that we don't waste computing time on formatting the log message when debug logging isn't turned on

src/prisma/cli/_node.py
RobertCraigie: Good point, I'll update this, thanks!

src/prisma/cli/_node.py
RobertCraigie: Yeah the unused ignores setting should be turned on, I think the issue may be that we installing an outdated version

src/prisma/cli/_node.py
jacobdr: Yeah I saw that it was a good catch. Regardless of where you add it, will be useful in bug reports we get

src/prisma/generator/models.py
RobertCraigie: We would probably be able to use an enum as a key but we do not currently have the architecture that resolves path names properly. I also think there may be cases where the generated platform name isn't always limited to a certain set of enum values.

Added documentation to this in a785d7c49e6a05ed04994fe81d1c3d4fa1aacb88.

src/prisma/_config.py
RobertCraigie: Changed the default cache directory to `~/.cache/prisma-python`, c1e76637845ae731a5ef8485fb98878dfeacb1b1.

PR #491, opened 2022-09-17
fix(package): force utf-8 in setup

tyteen4a03:
## Change Summary

Changes `setup.py` to always open files in UTF-8 encoding to avoid weird encoding errors when installing the package.

## Checklist

N/A

## Agreement


RobertCraigie:
Thank you! 💜

PR #512, opened 2022-10-06
fix(client): prevent the deletion of sqlite dbname

dv1x3r:
## Summary

I adjusted how SQLite URL is trimming "file:" and "sqlite:" prefixes in the beginning of connection string.
This was causing incorrect trimming when database file name starts with any character from those prefixes.

Example: "sqlite:teh.db" has been trimmed to "h.db".
The nature of lstrip method trims characters left to right while next symbol is found in the set of given characters.

Workaround: use relative path syntax like "sqlite:./teh.db".

## Checklist

## Agreement

### DIFF
--- a/src/prisma/generator/templates/client.py.jinja
+++ b/src/prisma/generator/templates/client.py.jinja
@@ -10,6 +10,7 @@ from ._types import BaseModelT
 from .engine import AbstractEngine, QueryEngine
 from .builder import QueryBuilder
 from .generator.models import EngineType, OptionalValueFromEnvVar
+from ._compat import removeprefix
 
 
 __all__ = (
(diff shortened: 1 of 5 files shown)


src/prisma/generator/templates/client.py.jinja
```diff
@@ -347,7 +347,7 @@ class Prisma:
         }
 
     def _make_sqlite_url(self, url: str, *, relative_to: Path = SCHEMA_PATH.parent) -> str:
-        url_path = url.lstrip('file:').lstrip('sqlite:')
+        url_path = url.lstrip('file').lstrip('sqlite')[1:]
```
RobertCraigie, line 350: I think the best way to fix this would be to just replace `.lstrip` with `.removeprefix` but since `removeprefix` was only added in `3.9` we'll have to define our own function that does this, e.g.

```py
def removeprefix(string: str, prefix: str) -> str:
    if string.startswith(prefix):
        return string[len(prefix):]
    return string
```

Could you add this function to `_compat.py` and use it here please? :)

RobertCraigie (commented):
Thanks for catching and reporting this bug!

Could you also please add a test case that demonstrates the bug as well, thanks!

RobertCraigie:
Thank you @dv1x3r!

Issue #507, opened 2022-10-01
Add support for `binaryTargets` generator option


RobertCraigie:
## Problem

As of #454 we do not properly generate the current binary platform name. This is fine for now as Prisma gives us the binary paths at generation time but because we cannot properly detect which binary we have to use we have to iterate through all of them to find one that works, this is very hacky.

## Suggested solution

We should copy the implementation from https://github.com/prisma/prisma/tree/main/packages/get-platform.

jacobdr:
* `prisma-client-py` version 7.0.0

Hitting this issue not due to the linked PR but due to something with docker cross-platform build on OSX M1 for `--platform=linux/amd64`. 

Certainly the `binaryTargets` is not supported, but I also did not see a warning message as I would have expected given [this section of code](https://github.com/RobertCraigie/prisma-client-py/blob/0df8824a8099e2f39079684fb136b23000ccbf72/src/prisma/generator/models.py#L371), through arguably it might never get invoked since the runtime error around the engine (`qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory`) happens before this gets executed.

```bash
(arcadia-py3.10) arcadia-user@182b77edc2fa:~/arcadia$ uname -m
aarch64
(arcadia-py3.10) arcadia-user@182b77edc2fa:~/arcadia$ uname -a
Linux 182b77edc2fa 5.10.124-linuxkit #1 SMP PREEMPT Thu Jun 30 08:18:26 UTC 2022 aarch64 GNU/Linux
(arcadia-py3.10) arcadia-user@182b77edc2fa:~/arcadia$ file /tmp/prisma/binaries/engines/efdf9b1183dddfd4258cd181a72125755215ab7b/prisma-migration-engine-debian-openssl-1.1.x
/tmp/prisma/binaries/engines/efdf9b1183dddfd4258cd181a72125755215ab7b/prisma-migration-engine-debian-openssl-1.1.x: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, BuildID[sha1]=ed74130fc99ac614964ce0e7f711c4583bced8c3, with debug_info, not stripped
```

More debugging context:

```
(arcadia-py3.10) arcadia-user@182b77edc2fa:~/arcadia$ prisma py fetch
Downloading binaries  [####################################]  100%
[...]

PR #521, opened 2022-10-16
feat(ci): Add initial test case for a Docker build in CI for multi-platform builds

jacobdr:
## Change Summary

This PR adds a basic docker build setup that exercises the `prisma` CLI, building for multiple platforms in Docker via `--platform=linux/amd64,linux=arm64`

Additionally, it starts to exercise the build against multiple [Linux distributions for Python](https://hub.docker.com/_/python), starting with a basic support matrix for Debian Buster and Alpine (latest).

The scope is **not to publish** any sort of official image, but rather since GH Actions does not support an ARM CPU runner we can emulate how a build would have performed via the QEMU emulation of docker.

Relates to #454
Relates to #461
Relates to #507 

CI results for this running off of my fork can be found here: https://github.com/jacobdr/prisma-client-py/pull/1

## Checklist

## Agreement

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -255,7 +255,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        docker-platform: ["linux/amd64", "linux/arm64"]
+        docker-platform: ["linux/amd64"]
+        # TODO: Uncomment this to add testing support for arm64 and delete
+        # the above
+        # docker-platform: ["linux/amd64", "linux/arm64"]
         python-os-distro: ["slim-bullseye", "alpine"]
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


RobertCraigie:
Love this change, I should've done this ages ago, thanks!

This looks good to me but I'd prefer if it was more obvious that the `Dockerfile` is only used for testing purposes. I'm not sure where it would best fit into the current architecture though...

jacobdr:
I could place it in the `tests` directory so its a bit more hidden and leave a comment up at the top of the file indicating why. Would that be sufficient?

.github/workflows/test.yml
```diff
@@ -248,3 +248,31 @@ jobs:
...
+  docker:
+    name: docker
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        docker-platform: ["linux/amd64"]
+        # TODO: Uncomment this to add testing support for arm64 and delete
+        # the above
+        # docker-platform: ["linux/amd64", "linux/arm64"]
```
RobertCraigie, line 261: Is `linux/arm64` currently broken?

RobertCraigie: Just saw your comment in another file, what was the error?

RobertCraigie (approved):
All looks good to me, thanks!

PR #525, opened 2022-10-22
feat(ci): Add docker windows test

RobertCraigie:
## Change Summary

## Checklist

## Agreement


jacobdr:
@RobertCraigie feel free to add me as a reviewer in the future on Docker stuff. The duplication of the dockerfile here I think we might want to patch up in the future for easier maintenance (but not a big deal, Windows is admittedly always going to be a different beast so perhaps this split was inevitable)

RobertCraigie:
@jacobdr Will do :)

Yeah I thought it best to keep it separate as I'm not *that* familiar with docker so running conditional commands or anything else that would be needed to make it work in the same Dockerfile didn't seem to be worth it to me, it also just might be worse to maintain.

PR #526, opened 2022-10-23
chore(project): Removed UnicodeExpr as it is no longer supported

rhoboro:
## Change Summary

`mypy` command gives the following error and I found https://github.com/python/mypy/pull/13254 and https://github.com/python/mypy/issues/12237 .
So, I removed it.

Screen Shot 2022-10-23 at 14 31 21

## Checklist

## Agreement

### DIFF
--- a/src/prisma/mypy.py
+++ b/src/prisma/mypy.py
@@ -30,7 +30,6 @@ from mypy.nodes import (
     NameExpr,
     Var,
     BytesExpr,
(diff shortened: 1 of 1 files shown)


RobertCraigie:
Thanks for fixing this!

It looks like `UnicodeExpr` was actually only used for Python 2 which I had not realised when originally writing the mypy plugin.

Issue #266, opened 2022-02-03
Add tests for all supported database providers


RobertCraigie:
## Problem

We currently do not run integration tests on all of the database providers we support as Prisma tests them themselves, however we should still run our own tests as there may be database specific functionality that causes errors with our client.

## Suggested solution

Add new integration tests for all supported database providers: https://www.prisma.io/docs/reference/database-reference/supported-databases

## Additional context

For potential contributors: https://prisma-client-py.readthedocs.io/en/stable/contributing/contributing/#integration-tests

Please only submit pull requests with *one* database provider at a time. Thank you :)

Issue #35, opened 2021-07-15
Installing from a local package results in the new package being polluted


RobertCraigie:
## Bug description

Installing from a local package that has been generated results in the new installation being polluted by the previous installation.

## How to reproduce

This is most easily reproducible by mixing virtual environments and http dependencies.

```shell
git clone https://github.com/RobertCraigie/prisma-client-py prisma-client-py
cd prisma-client-py
python3 -m venv .venv && source .venv/bin/activate
pip install -U -e .[aiohttp]
prisma generate
deactivate
mkdir sync && cd sync
python3 -m venv .venv && source .venv/bin/activate
pip install -U ..[requests]
prisma --help  # import error: aiohttp
```
```prisma
datasource db {
  provider = "sqlite"
  url      = "file:dev.db"
}

generator db {
  provider = "prisma-client-py"
  http     = "aiohttp"
}

model User {
  id Int @id
}
```

RobertCraigie:
I do not think this is solvable by us.

Installing from a locally generated package is not a standard or supported use case. If this is required then a cleanup script can be ran before installing.

PR #528, opened 2022-10-24
Simplify the requirements setup for integration tests

jacobdr:
## Change Summary

I am going to start working on #266, but in starting that I realized we had a lot of duplication in the integration test `test.sh` entrypoints. This PR tries to simplify our setup by harmonzing the integration test virtual environments, making it easier to update dependencies and maintain common requirements. 

The downside of this approach is that each virtualenv might have _more_ dependencies than it needs which could increase installation time. However, in practice this is trivial, and I think the maintainability improvement trumps the speed/isolation concerns.

- [ ] Check to see if this removes the workaround introduced in #35 in `tests/integrations/sync/test.sh`

## Checklist

## Agreement


jacobdr:
@RobertCraigie I think we could push this further get rid of the `requirements/coverage.txt` and `requirements/dev.txt` and really keep everything in one place for "developers"

jacobdr:
@RobertCraigie is this something that I should continue on? The base problem is that as a dev getting the IDE to work across all the different environments we have is pretty taxing, though admittedly I might be missing something. The general idea is to have one set of "runtime dependencies" and another set of "dev dependencies" and that's it. 

The goal: One pip/poetry install in CI (or local) of the dev dependencies should give me a fully functioning environment

jacobdr:
If you wanted to separate them further I think we would want to switch to [poetry dependency groups](https://python-poetry.org/docs/master/managing-dependencies/#dependency-groups) -- which I think would get around the issue you saw in #35 and allow us to generate a lockfile for dev dependencies to get a more stable install. The only downside I see there is potentially with `nox` but we could cross that bridge if indeed it proves an issue

PR #529, opened 2022-10-26
refactor(tests): Pass through coverage context, refactor download binary tests to use mocks

jacobdr:
## Change Summary

This MR is the result of my local installation and having run the test suite now many times. It does a few things to help make local development easier:

* Adds the a [static context](https://coverage.readthedocs.io/en/6.4.4/contexts.html) to the coverage so that we know which Python versions executed which sections of code. This will be eve more useful once we add the integration tests for different databases
* The `test_fetch` module which exercises the CLI was running notoriously slow on my machine. I am in the US and it seems Prisma hosts the binaries in an EU-based S3, which should have still been fast, but was not. Regardless, I turned this into more of a pure unit test with mocking, but left a real "download" test in `tests/integrations/sync/tests/test_binary.py`
* Speaking of `tests/integrations/sync/tests/test_binary.py`, I switched that to download the smallest Prisma binary (the formatter) so as to spare long download times
* Add the longest test durations to the `pytest` output to we know which tests are taking the longest
* Added a global test timeout of 3 minutes (applies to each test function) to help capture issues as people add things that might get quite slow in CI (this helped me identify the issues with the fetch tests)

The one thing that is slightly suspect about this changeset are the changes I had to make to the `_maybe_warn_subclassing` function. I could not for the life of me get that to work locally on my machine in Python 3.7. I think this solution is slightly more robust, but also slightly slower. Please definitely take a look at that change.

## Checklist

- [X] Unit tests for the changes exist

## Agreement

### DIFF
--- a/tests/test_generation/exhaustive/__snapshots__/test_exhaustive/test_sync[models.py].raw
+++ b/tests/test_generation/exhaustive/__snapshots__/test_exhaustive/test_sync[models.py].raw
@@ -66,17 +66,27 @@ _implicit_subclass_packages: Set[str] = {
     'fastapi',
 }
 
+def _module_to_parent_package(frame):
+    try:
+        module_name = inspect.getmodule(frame).__name__
+        parent_package, *_ = module_name.split(".")
+        return parent_package
+    except:
+        return None
 
-def _maybe_warn_subclassing(new_model: str, base_model: str, *, stacklevel: int = 3) -> None:
+def _maybe_warn_subclassing(new_model: str, base_model: str, *, max_call_stack_depth_to_inspect: int = 5) -> None:
     # at least 3 frames are guaranteed to exist if we are being called from __init_subclass__
     # stack: 1 = __init_subclass__, 2 = abc, 3 = <caller>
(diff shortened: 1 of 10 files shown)


tests/test_generation/exhaustive/__snapshots__/test_exhaustive/test_sync[models.py].raw
```diff
@@ -66,17 +66,27 @@ _implicit_subclass_packages: Set[str] = {
...
+def _module_to_parent_package(frame):
+    try:
+        module_name = inspect.getmodule(frame).__name__
+        parent_package, *_ = module_name.split(".")
+        return parent_package
+    except:
+        return None
 
-def _maybe_warn_subclassing(new_model: str, base_model: str, *, stacklevel: int = 3) -> None:
+def _maybe_warn_subclassing(new_model: str, base_model: str, *, max_call_stack_depth_to_inspect: int = 5) -> None:
```
Comment (jacobdr) [inline tests/test_generation/exhaustive/__snapshots__/test_exhaustive/test_sync[models.py].raw:77]: This is the sketchiest change I had to make to get the test suite to run locally. Needs you eyeballs @RobertCraigie

PR #538, opened 2022-10-29
fix(ci): Docker build issues

jacobdr:
## Change Summary

* Fix issue with testing alpine-based docker in CI. However, since the Go Prisma CLI does not work on Alpine, I had to disable the test as it fails
* Start to set up PKI machinery for Windows to deal with nodeenv downloads -- unclear why we need this but starts to put the infrasctructure there inc ase we need it
* Implement better caching to speed up docker build times -- but does not work on Windows since we don't have buildx so we won't see an improvement there

In the initial PR (#521) where I added Docker-based CI I forgot to pass through the Docker build arg with `=` -- so the variation we wanted to test on Alpine was not working as expected. This PR fixes that, so that the `OS_DISTRO` is properly populated

## Checklist

- [X] Unit tests for the changes exist
- [X] Tests pass without significant drop in coverage

## Agreement

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -289,5 +289,8 @@ jobs:
       - uses: actions/checkout@v3
       - name: Docker Build
         # Use --% to allow double hyphen
+        # Caching not currently working since we don't use buildx yet, windows
+        # support seems poor because of interactions with --prvileged
+        # --cache-from=type=gha --cache-to=type=gha,mode=max
         run: |
-          docker build --% -f tests/windows.Dockerfile -t prisma-client-py --cache-from=type=gha --cache-to=type=gha,mode=max .
+          docker build --% -f tests/windows.Dockerfile -t prisma-client-py .
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -274,7 +274,11 @@ jobs:
           file: tests/Dockerfile
           platforms: "${{ matrix.docker-platform }}"
           build-args: OS_DISTRO=${{ matrix.python-os-distro }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
 
+  # Note: Windows does not play well with the standard GH Docker + buildx
+  # actions
   docker-windows:
     name: docker (windows)
     runs-on: windows-latest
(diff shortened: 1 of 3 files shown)


jacobdr:
@RobertCraigie This is now ready to review. The idea is for this to merge to main and then you can re-base your CLI PRs off of this guy so we can see what the new support matrix looks like

.github/workflows/test.yml
```diff
@@ -275,14 +276,21 @@ jobs:
...
   docker-windows:
     name: docker (windows)
     runs-on: windows-latest
     steps:
       - uses: actions/checkout@v3
-
       - name: Docker Build
+        # Use --% to allow double hyphen
+        # Caching not currently working since we don't use buildx yet, windows
+        # support seems poor because of interactions with --prvileged
```
RobertCraigie, line 293:
```suggestion
        # support seems poor because of interactions with --privileged
```

tests/Dockerfile
```diff
@@ -6,15 +6,50 @@ ARG OS_DISTRO=slim-bullseye
...
+RUN \
+    if [[ $OS_DISTRO =~ alpine ]]; then \
+    adduser -u ${PRISMA_USER_ID} -D prisma; \
+    apk add gcc musl-dev yaml-dev yaml libffi-dev; \
+    elif bash -c "[[ ${OS_DISTRO} =~ slim ]]"; then \
+    useradd --create-home --uid ${PRISMA_USER_ID} --shell /bin/bash prisma; \
+    else \
+    echo "Unrecognized distro $OS_DISTRO"; \
+    exit 99; \
+    fi
```
RobertCraigie, line 43: Minor nit, I think this makes it a lot more readable (this is valid syntax right?)

```suggestion
    if [[ $OS_DISTRO =~ alpine ]]; then \
      adduser -u ${PRISMA_USER_ID} -D prisma; \
      apk add gcc musl-dev yaml-dev yaml libffi-dev; \
    elif bash -c "[[ ${OS_DISTRO} =~ slim ]]"; then \
      useradd --create-home --uid ${PRISMA_USER_ID} --shell /bin/bash prisma; \
    else \
      echo "Unrecognized distro $OS_DISTRO"; \
      exit 99; \
    fi
```

RobertCraigie (approved):
Thanks for figuring the alpine issues out and for the whole mess with Windows!

PR #542, opened 2022-10-31
feat(tests): add database testing suite

### LINKED ISSUE #266: Add tests for all supported database providers

RobertCraigie:
## Change Summary

Closes #266.

TODO:
- [x] Update contributing documentation
- [x] Run tests under coverage

## Checklist

## Agreement


.github/workflows/test.yml
```diff
@@ -239,9 +239,60 @@ jobs:
...
+          pip install -U -r pipelines/requirements/ci.txt .[dev]
+
+      - uses: actions/cache@v3
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('requirements/*.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+
+      # TODO: cache prisma engines
```
jacobdr, line 284: That would really help with the CI time, but feel free to spin this out into another ticket and I would be happy to pick it up. I think you had a running TODO list ticket to chuck it into

databases/tests/conftest.py
```diff
@@ -0,0 +1,101 @@
...
+from lib.testing import async_fixture
+from .utils import RAW_QUERIES_MAPPING, RawQueries
+from ..utils import DatabaseConfig
+
+
+if TYPE_CHECKING:
+    from _pytest.fixtures import FixtureRequest
+
+
+# TODO: a lot of this is copied from `tests/conftest.py`, figure out a good way to share
```
jacobdr, line 22: Add this to the TODO tracker. Is it as simple as imports from conftest? It might be a little bit an an anti-pattern but I think the value of code re-use trumps

jacobdr:
@RobertCraigie this was really good work m8

jacobdr:
I'm kind of surprised by the coverage hit -- I would have expected it to be _higher_ after this PR. Do you know what the deal is with that?

RobertCraigie:
Thanks!

RobertCraigie:
@jacobdr I think the coverage drop is partly because the actual CLI script behind the test suite is not being picked up by coverage correctly, aside from that it looks like codecov are including some files that have been deleted.

.github/workflows/test.yml
RobertCraigie: Good point - #550.

databases/tests/conftest.py
RobertCraigie: Good idea, I think that sort of thing is a bit weird but works really well for this case, done in cbddf0b6f2070c48e9fdd5b158b4dcb2fdbd3f2e

PR #678, opened 2023-01-20
fix(client): ensure SIGINT is unblocked before forking query engine

ezorita:
## Change Summary

Forking of client engine query processes relies on `SIGINT` signal to disconnect. We must ensure this signal is unblocked before forking the process, otherwise `process.wait()` in disconnect will end up in a deadlock.

I discovered this issue while running `prisma` on `ray`. It seems `ray` workers inherit a blocking mask (they shouldn't!) and that would cause `prisma` to end in a deadlock each time it tries to disconnect the engine.

## Checklist

## Agreement


RobertCraigie:
Interesting, thanks! I was completely unaware of this even being a potential issue. I'll need to do some research on this before being able to review it properly to fully understand the effects of this change.

I'm also thinking based off of the problem statement we should add a timeout to `proces.wait()`, would that help in this situation? I'm unsure what the timeout should actually be though.

RobertCraigie:
Also this seems like a very good reason to fix the integration tests and have some e2e treating setup in place so we can test integrations with third party libraries!

(Not suggesting that would have to be done in this PR, just thinking out loud)

ezorita:
> Interesting, thanks! I was completely unaware of this even being a potential issue. I'll need to do some research on this before being able to review it properly to fully understand the effects of this change.
> 
> I'm also thinking based off of the problem statement we should add a timeout to `proces.wait()`, would that help in this situation? I'm unsure what the timeout should actually be though.

Absolutely, take your time to review this in depth.

My first approach was to add a timemout to `process.wait()` but that would make the library terribly slow in case `SIGINT` is blocked in the child process, e.g. each call to `disconnect()` would block for at least `timeout` seconds. Said that, I agree both solutions would work better together; in case something weird happened to the child process or if it blocked the signal itself, exhausting the timeout and issuing a `SIGKILL` (can't be blocked) would do the trick:

```python
    def close(self, timeout: int = 1) -> None:
        log.debug('Disconnecting query engine...')

        if self.process is not None:
            if platform.name() == 'windows':
                self.process.kill()
                self.process.wait()
            else:
                self.process.send_signal(signal.SIGINT)
               try:
                   self.process.wait(timeout=timeout)
               except subprocess.TimeoutExpired:
                   self.process.send_signal(signal.SIGKILL)

            self.process = None

        {% if not is_async %}
        self._close_session()
        {% endif %}
        log.debug('Disconnected query engine')

```
A more brutal approach would be to send a `SIGKILL` directly but that won't let the child process to terminate gracefully.

ezorita:
@RobertCraigie I updated tests and applied some fixes, but still I can't seem to get rid of this error in the database tests:
```
 /home/runner/work/prisma-client-py/prisma-client-py/.nox/databases/lib/python3.10/site-packages/prisma/engine/query.py
  /home/runner/work/prisma-client-py/prisma-client-py/.nox/databases/lib/python3.10/site-packages/prisma/engine/query.py:181:24 - error: Cannot assign member "process" for type "QueryEngine"
    Type "Popen[str]" cannot be assigned to type "Popen[bytes] | None"
      "Popen[str]" is incompatible with "Popen[bytes]"
        TypeVar "AnyStr@Popen" is invariant
          "str" is incompatible with "bytes"
      Type cannot be assigned to type "None" (reportGeneralTypeIssues)
1 error, 0 warnings, 0 informations 
```
For some reason the call to `subprocess.Popen()` returns a `Popen[str]` object, even though I made `text=False` explicit in the `Popen()` call.

RobertCraigie:
> My first approach was to add a timemout to process.wait() but that would make the library terribly slow in case SIGINT is blocked in the child process, e.g. each call to disconnect() would block for at least timeout seconds. Said that, I agree both solutions would work better together

Yeah I agree it looks like both solutions would work well together.

> For some reason the call to subprocess.Popen() returns a Popen[str] object, even though I made text=False explicit in the Popen() call.

Yeah that seems weird, you probably need to specify a different argument to get the overloads to work properly but I'm not sure...

If it would be easier you could change the type of the `process` property to `Popen[str | bytes]` which should probably work. I don't think there are any cases where we explicitly handle the output of the process so I don't think it matters.

ezorita:
> If it would be easier you could change the type of the `process` property to `Popen[str | bytes]` which should probably work. I don't think there are any cases where we explicitly handle the output of the process so I don't think it matters.

Awesome, I had this fix in mind but wanted to make sure using `subprocess.Popen[str]` won't break anything.

ezorita:
@RobertCraigie are you planning to move this forward for the next release? No rush, just to know, otherwise I will apply a manual patch when installing the lib in my stack.

Thanks!

src/prisma/generator/templates/engine/query.py.jinja
```diff
@@ -132,11 +136,20 @@ class QueryEngine(HTTPEngine):
...
+        popen_kwargs: Dict[str,Any] = {
+            "env": env,
+            "stdout": sys.stdout,
+            "stderr": sys.stderr,
+            "text": False,
+        }
+        if platform.name() != 'windows':
+            popen_kwargs["preexec_fn"] = lambda: signal.pthread_sigmask(
+                signal.SIG_UNBLOCK, [signal.SIGINT, signal.SIGTERM]
+            )
```
RobertCraigie, line 148: So in the documentation there is a [big warning](https://docs.python.org/3/library/subprocess.html#:~:text=The%20preexec_fn%20parameter%20is%20NOT%20SAFE%20to%20use%20in%20the%20presence%20of%20threads%20in%20your%20application.%20The%20child%20process%20could%20deadlock%20before%20exec%20is%20called.) that this is not safe to use alongside threads, does this matter in our case?

I would think it doesn't right? As in our case we're just using it as an attempt to avoid deadlocks and if it deadlocks before `preexec_fn` is even called then there's nothing we can do, am I understanding this correctly?

src/prisma/generator/templates/engine/query.py.jinja
```diff
@@ -46,16 +46,20 @@ class QueryEngine(HTTPEngine):
     def __del__(self) -> None:
         self.stop()
 
-    def close(self) -> None:
+    def close(self, timeout: float = 1.0) -> None:
```
RobertCraigie, line 49: I'm not sure if this is the right timeout but I also don't have a better suggestion....

I think for now it might be best to make this optional and default to `None` and support passing this through from `db.disconnect()` so you can set the default to `1` if that makes sense in your application.

RobertCraigie:
@ezorita Yeah I am planning on pushing a new release over the weekend and I'll try to get this merged before then :)

src/prisma/generator/templates/engine/query.py.jinja
ezorita: Correct. One could deadlock a child process when blocking signals, but I don't see how a deadlock could be produced when unblocking them. To my understanding this would be specific to multithreading environments since blocked signals may cause thread synchronization issues, but in our case we're just forking a process.

src/prisma/generator/templates/engine/query.py.jinja
ezorita: Makes total sense. I will also revert the `Popen[str]` type definition since it is still failing... I will try setting it to `Popen[str]` everywhere...

ezorita:
I have a question about the generator for `client.py`, the disconnect function:
https://github.com/RobertCraigie/prisma-client-py/blob/f5ea308c6b7df96d6707e2b108b9a5ae78e32595/src/prisma/generator/templates/client.py.jinja#L235
```jinja2
    {{ maybe_async_def }}disconnect(self) -> None:
        """Disconnect the Prisma query engine."""
        if self.__engine is not None:
            {% if is_async %}
            self.__engine.close()
            {{ maybe_await }}self.__engine.aclose()
            {% else %}
            self.__engine.stop()
            {% endif %}
            self.__engine = None
```
Shouldn't the client call `.stop()` no matter what? It seems to me that the async branch is calling close twice, but we're not closing the engine in sync mode.

Suggested update:
```jinja2
    {{ maybe_async_def }}disconnect(self) -> None:
        """Disconnect the Prisma query engine."""
        if self.__engine is not None:
            {% if is_async %}
            {{ maybe_await }}self.__engine.aclose()
            {% else %}
            self.__engine.close()
            {% endif %}
            self.__engine.stop()
            self.__engine = None
```

RobertCraigie:
No I think the current behaviour is correct. Admittedly this is an awkward internal API design...

What `stop()` does is just call `close()` and create an async task to call `aclose()`. I definitely think these methods could be named / documented better, sorry about this...

RobertCraigie (approved):
Looks good to me! Thanks for working on this :)

Sorry it's a week later than I had thought.

Going to merge as soon as CI passes.

ezorita:
Thank you @RobertCraigie, this is a beautiful project!

PR #823, opened 2023-10-02
feat(client) - decorator to provide auto-connecting and auto-disconnecting Prisma client

LucaTabone:
## Change Summary

Added a decorator **use_prisma** which provides an auto-connecting and auto-disconnecting prisma client to the function it decorates. It is a simple way to use the Prisma client.

## Checklist

## Agreement

### DIFF
--- a/src/prisma/decorator.py
+++ b/src/prisma/decorator.py
@@ -1,0 +1,66 @@
+import asyncio
+from prisma import Prisma
+from functools import wraps
+from datetime import timedelta
+from typing import Any, Callable
+
+
+__all__ = ('use_prisma',)
+
+CLIENT_NAME_DEFAULT_VALUE: str = 'db'
+
+
+def use_prisma(*args: Any, **kwargs: Any) -> Callable[..., Any]:
+    """Decorator to provide an auto-connecting and auto-disconnecting Prisma client."""
+    _CLIENT_NAME_KEY: str = 'name'
+
+    def get_client_settings():
+        """Return the appropriate settings based on provided arguments."""
+        settings = {  # default client settings
+            'connect_timeout': timedelta(seconds=30),
+            'http': {'timeout': 600},
+        }
+        if not kwargs:
+            return settings
+        settings.update(kwargs)
+        if _CLIENT_NAME_KEY in settings:
+            del settings[_CLIENT_NAME_KEY]
+        return settings
+
+    def should_bypass(*f_args, **f_kwargs):
+        """
+        Check if the decorated function was called with a Prisma instance in the args or kwargs.
+        If yes, return True, otherwise return False.
+        In case of yes, we won't create a new Prisma instance.
+        """
+        for arg_list in [f_args, f_kwargs.values()]:
+            for arg in arg_list:
+                if isinstance(arg, Prisma):
+                    return True
+        return False
+
+    def outer_wrapper(func):
+        @wraps(func)
+        async def wrapper(*f_args, **f_kwargs):
+            if should_bypass(*f_args, **f_kwargs):
+                return await func(*f_args, **f_kwargs)
+            prisma = Prisma(**get_client_settings())
+            await prisma.connect()
+            try:
+                f_kwargs[
+                    kwargs.get(_CLIENT_NAME_KEY, CLIENT_NAME_DEFAULT_VALUE)
+                ] = prisma
+                return await func(*f_args, **f_kwargs)
+            except (KeyboardInterrupt, asyncio.CancelledError):
+                await prisma.disconnect()
+            finally:
+                await prisma.disconnect()
+
+        return wrapper
+
+    # distinguish between decorator usage with and without arguments
+    if args and callable(args[0]):
+        # no decorator arguments were provided
+        return outer_wrapper(args[0])
+    # decorator arguments were provided
+    return outer_wrapper
--- a/tests/test_decorator.py
+++ b/tests/test_decorator.py
@@ -23,8 +23,10 @@ async def test_use_prisma_decorator():
     await _create_user()
     await _get_user()
     await _delete_user()
-    await _with_client_name_arg()
-    await _with_connect_timeout_arg()
+    await _with_custom_client_name()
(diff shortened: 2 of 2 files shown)


src/prisma/decorator.py
```diff
@@ -0,0 +1,66 @@
...
+from datetime import timedelta
+from typing import Any, Callable
+
+
+__all__ = ('use_prisma',)
+
+CLIENT_NAME_DEFAULT_VALUE: str = 'db'
+
+
+def use_prisma(*args: Any, **kwargs: Any) -> Callable[..., Any]:
```
kkom, line 13: Hey @LucaTabone – I'm just passing by here and have one thing to comment on – hope that's helpful!

Using `Any` and `...` in the type signature of a decorator obscures the types of the annotated function, which can be dangerous. You can use the generic [ParamSpec](https://docs.python.org/3/library/typing.html#typing.ParamSpec) to avoid this problem.

PS: will leave to @RobertCraigie whether this makes sense as a first class part of the library

LucaTabone: Hey @kkom :) Thanks a lot for your comment!

I agree! Unfortunately ParamSpec is available for Python 3.10+, but the tests within this repository test against 3.9 too.

Is there another way you see on how we could provide better type-hinting here in light of Python 3.9?

kkom: Oh, I see! I don't have much experience creating Python libraries, just applications, but would https://pypi.org/project/typing-extensions/ help here?

We got the "@override" decorator in Python 3.11 in this way just now.

RobertCraigie:
Can you share more about the intended use case for this decorator?

I'm wary of including something like this in the core library because connecting the client can be a fairly expensive operation and it would be easy for someone to overuse the decorator without fully realising the ramifications of it. I've already seen people use the context manager for connecting / disconnecting directly in their server code without fully realising that it was slowing their application down.

LucaTabone:
Hey Robert,

Thanks so much for taking the time to review my PR. I understand your concerns, and they're absolutely valid. Let me shed some light on the intended use cases and the potential advantages of this decorator:

**Ergonomics & Developer Experience**: For many developers, especially those who are new to Prisma or database operations in general, the connection management part can be intimidating. This decorator aims to abstract away some of that complexity, allowing them to focus on the logic of their database operations.

**Error Handling and Ensuring Closure**: One of the challenges developers often face is ensuring that connections are properly closed, especially in scenarios where there might be exceptions or unexpected terminations. The decorator ensures that the Prisma client is gracefully disconnected, even if there's an error in the decorated function (in the _finally_ block we disconnect the prisma client).

**Explicitness Over Implicitness**: By providing the prisma client as a parameter to the decorated function, it makes the use of the Prisma client within that function explicit. This encourages developers to think about where they're using the client and why.

**Education**: I absolutely agree that misuse can lead to performance degradation. I believe this can be mitigated with proper documentation and examples highlighting both the correct use cases and potential pitfalls. We could even add warnings in the doc-strings or documentation to inform users about potential performance issues if used carelessly.

**Optional Utility**: It's important to remember that this is just a utility for those who find it beneficial. Advanced users or those who have specific connection management requirements can always bypass the decorator and manage connections in their preferred way.

Thanks for considering this, and I look forward to your feedback!

RobertCraigie:
Thanks @LucaTabone, those are all great points!

Are you aware that we provide a context manager as well? I think that the majority of your points also apply to the context manager.

```py
with Prisma() as db:
  await db.users.create(...)
```

This is explicit and automatically handles errors for you. I appreciate that context managers can be intimidating for newer python users but so can decorators imo. Thoughts?

LucaTabone:
Thanks for pointing that out! You're right, the context manager is quite handy for managing the client connections. It's straightforward, and does the job well with error handling.

I tossed in the decorator idea as a kind of "quick apply" feature. Here's why I think it could be a cool buddy to the context manager rather than a competitor:

**Ease of use**
Just slap @use_prisma above a function, and voilà - it's Prisma-ready. It feels like a small thing, but it's a time-saver when you're juggling a bunch of functions.

**Keeping things tidy**
The decorator cuts down on nesting. We all know how Python code can turn into an unintended Christmas tree with all the indentations!

**Readability perks**
It also acts like a neon sign saying, "Hey, this function talks to the database!" - making life a bit easier when you're skimming through code.

All in all I'm just thinking of giving folks another tool in their belt for different scenarios. No one-size-fits-all, right?

Keen to hear your thoughts on this!

LucaTabone:
Hey @RobertCraigie, is there any more info I can provide here?

kkom:
Just to add my 0.02 USD here – I'm not a maintainer, but I'm following this project. I would side with @RobertCraigie's view here – I think there should be a very high bar to adding helpers and additional ways of doing things to libraries like Prisma Python.

They add maintanance work, and if unmaintained – can cause problems for users in the future. The problem with typing in https://github.com/RobertCraigie/prisma-client-py/pull/823#discussion_r1349551998 is one example. There are also other things - like default client name or passing-through settings. I see 2 settings are set in this new decorator – what if some important settings are added and forgotten to port over to this methods, etc.

LucaTabone:
> Just to add my 0.02 USD here – I'm not a maintainer, but I'm following this project. I would side with @RobertCraigie's view here – I think there should be a very high bar to adding helpers and additional ways of doing things to libraries like Prisma Python.
> 
> They add maintanance work, and if unmaintained – can cause problems for users in the future. The problem with typing in [#823 (comment)](https://github.com/RobertCraigie/prisma-client-py/pull/823#discussion_r1349551998) is one example. There are also other things - like default client name or passing-through settings. I see 2 settings are set in this new decorator – what if some important settings are added and forgotten to port over to this methods, etc.

Hey @kkom, Thanks for your reply here!
I really appreciate you taking the time to think about this new feature.

I feel like your points are definitely valid and apply to all open source projects. Every new feature introduces maintenance.
Regarding default arguments, where do you see this impacting dev experience negatively?

How would you change the code in this PR in order to provide maximum value to devs?

Best,
Luca

PR #877, opened 2024-01-23
feat(generator): support custom default casing for client properties

lxxonx:
## Change Summary

```prisma
model UserProfile {
  ...
}
```
this code generate prisma class attributes like `prisma.userprofile`, which is sometimes very hard to read
i suggest having default attribute name as `prisma.user_profile` with using `jinja2_strcase` package.

## Checklist

- [  ] Documentation reflects changes where applicable

## Agreement


RobertCraigie:
Thanks! Unfortunately, this is a breaking change so I'd like to first introduce this using a config option, e.g.

```prisma
generator client {
  provider = "prisma-client-py"
  client_casing = "snake_case"
}
```

I'm also a little hesitant to introduce this new dependency, would it be feasible to do this without that? or at least using a popular package for changing case?

src/prisma/generator/templates/actions.py.jinja
```diff
@@ -159,7 +159,7 @@ class {{ model.name }}Actions(Generic[{{ ModelType }}]):
         -------
         ```py
         # create a {{ model.name }} record from just the required fields
-        {{ model.name.lower() }} = {{ maybe_await }}{{ model.name }}.prisma().create(
+        {{ model.name | to_snake | lower }} = {{ maybe_await }}{{ model.name }}.prisma().create(
```
RobertCraigie, line 162: note: I plan on changing how this works so that instead of doing `model.name.lower()` everywhere which makes changes like your one much more cumbersome.

In a separate PR I'll change this to `model.instance_name`.

src/prisma/generator/utils.py
```diff
@@ -122,3 +123,29 @@ def clean_multiline(string: str) -> str:
...
+        input_str = input_str[0].lower() + input_str[1:]
+    return UNDERSCORE_RE.sub(lambda m: m.group(0)[-1].upper(), input_str)
+
+
+def to_pascal_case(input_str: str) -> str:
+    def _replace_fn(match: re.Match[str]) -> str:
+        return match.group(1)[0].upper() + match.group(1)[1:]
+
+    input_str = to_camel_case(PASCAL_RE.sub(_replace_fn, input_str))
+    return input_str[0].upper() + input_str[1:] if len(input_str) != 0 else input_str
```
RobertCraigie, line 151: question: where did you get these functions from? did you come up with it yourself?

I'm only asking because, if you copied these from somewhere, it would be great to add a link to that place in case these need to be updated in the future.

src/prisma/generator/utils.py
```diff
@@ -122,3 +123,29 @@ def clean_multiline(string: str) -> str:
     assert string, 'Expected non-empty string'
     lines = string.splitlines()
     return '\n'.join([dedent(lines[0]), *lines[1:]])
+
+
+ACRONYM_RE = re.compile(r'([A-Z\d]+)(?=[A-Z\d]|$)')
+PASCAL_RE = re.compile(r'([^\-_]+)')
+SPLIT_RE = re.compile(r'([\-_]*[A-Z][^A-Z]*[\-_]*)')
+UNDERSCORE_RE = re.compile(r'(?<=[^\-_])[\-_]+[^\-_]')
```
RobertCraigie, line 131: nit: would be great to add test cases for all these functions!

src/prisma/generator/models.py
```diff
@@ -748,6 +758,13 @@ def instance_name(self) -> str:
...
         `User` -> `Prisma().user`
         """
+        config = get_config()
+        if isinstance(config, Config) and config.client_casing == ClientCasing.camel_case:
+            return to_camel_case(self.name)
+        elif isinstance(config, Config) and config.client_casing == ClientCasing.pascal_case:
+            return to_pascal_case(self.name)
+        elif isinstance(config, Config) and config.client_casing == ClientCasing.snake_case:
+            return to_snake_case(self.name)
         return self.name.lower()
```
RobertCraigie, line 768: nit: it would be great for this to exhaustively match the enum, you can do this by using our `assert_never()` helper function. Here's an example from another part of the codebase
```py
        if value == EngineType.binary:
            return value
        elif value == EngineType.dataproxy:  # pragma: no cover
            raise ValueError('Prisma Client Python does not support the Prisma Data Proxy yet.')
        elif value == EngineType.library:  # pragma: no cover
            raise ValueError('Prisma Client Python does not support native engine bindings yet.')
        else:  # pragma: no cover
            assert_never(value)
```

RobertCraigie (commented):
Haven't fully reviewed yet but love the direction!

RobertCraigie:
FYI I cherry picked your change case util functions into a separate PR as I needed them for something else, #918.

I tried to rebase your PR for you to resolve the merge conflicts but it looks like I don't have permission to push to your fork, could you open a separate PR with edit perms and from a different branch? (i.e. a branch that isn't `main` on your fork)

lxxonx:
New PR has opened with edit permission `https://github.com/RobertCraigie/prisma-client-py/pull/922`
Please check it out and let me know if you need anything for help :)

Issue #463, opened 2022-08-12
Add support for interactive transactions with an isolation level


RobertCraigie:
## Problem

Prisma recently added support for setting the isolation level for transactions. https://www.prisma.io/docs/concepts/components/prisma-client/transactions#transaction-isolation-level

Once we have support for transactions, #53, we should support this too.

## Suggested solution

Needs investigation.

PR #878, opened 2024-01-23
feat(client): add transaction isolation level

jonathanblade:
## Change Summary

Add support for interactive transactions with an isolation level.

Closes https://github.com/RobertCraigie/prisma-client-py/issues/463.

## Checklist

## Agreement


databases/tests/test_transactions.py
```diff
@@ -212,3 +212,129 @@ async def test_transaction_already_closed(client: Prisma) -> None:
...
+
+        non_repeatable_user = await tx1.user.find_first_or_raise(where={'id': user.id})
+        phantom_count = await tx1.user.count()
+
+    # Have dirty read
+    assert tx1_user.name != dirty_user.name
+    # Have non-repeatable read
+    assert tx1_user.name != non_repeatable_user.name
+    # Have phantom read
+    assert tx1_count != phantom_count
```
jonathanblade, line 244: @RobertCraigie Hi! I tried to add transaction isolation levels feature, but i don't understant, how to test it (tests not passes). I will be very grateful, if you help me find my mistake 🥺. I read a bit a theory and prototyped the following test workflow:
1. Create two different connections (`client` and `client2`).
2. Open transaction `tx1` in `client`.
3. Read some data in `tx1`.
4. Open transaction `tx2` in `client2`.
5. Modify the same data in `tx2`.
6. Read the same data in `tx1` before commit in `tx2` (test dirty read).
7. Read the same data in `tx1` after commit in `tx2` (test non-repeatable read and phantom read).

jonathanblade: I also tried to create tests like [here](https://github.com/prisma/prisma/blob/5.8.1/packages/client/tests/functional/batch-transaction-isolation-level/tests.ts#L49-L52), but even set `log_queries=True`, i don't see any transaction SQL queries.
And I don't understand the meaning of test like [here](https://github.com/prisma/prisma/blob/5.8.1/packages/client/tests/functional/interactive-transactions/tests.ts#L710-L725)...

RobertCraigie: Hey sorry for the delay in responding. I'm actually not very familiar with isolation levels myself, I'll look into it.

I've also caused merge conflicts for you (sorry!), I'll fix them.

RobertCraigie: > I also tried to create tests like [here](https://github.com/prisma/prisma/blob/5.8.1/packages/client/tests/functional/batch-transaction-isolation-level/tests.ts?rgh-link-date=2024-01-24T07%3A13%3A25Z#L49-L52), but even set log_queries=True, i don't see any transaction SQL queries.

Yeah unfortunately we don't have a good structured logging setup for the query engine like the JS client which makes things like this really difficult, I suspect they have some separate log level setup somewhere :shrug:

RobertCraigie: @jonathanblade what was the thinking behind creating a second client in the `test_read_uncommited_isolation_level` test?

jonathanblade: > @jonathanblade what was the thinking behind creating a second client in the `test_read_uncommited_isolation_level` test?

@RobertCraigie To be sure that transactions runs in different connections.

jonathanblade: > @jonathanblade what was the thinking behind creating a second client in the `test_read_uncommited_isolation_level` test?

Do you think the following is ok?
```python
async with client.tx(isolation_level=prisma.TransactionIsolationLevel.READ_UNCOMMITED) as tx1, client.tx() as tx2:
    # test workflow
```

RobertCraigie: >  To be sure that transactions runs in different connections.

Ah that makes sense, thanks for clarifying 👍

I'm not too sure what's happening with that test case as well, I spent some time looking into it and I can't see why it's not working as expected...

RobertCraigie: Actually, reading the psql docs on this, it looks like they make it difficult to have a dirty read - https://www.postgresql.org/docs/current/transaction-iso.html (only read the start so far).

I'm also seeing the same behaviour in the JS client

```ts
  const user = await prisma.user.create({ data: { name: "First" } });

  await prisma.$transaction(
    async (tx) => {
      await tx.user.update({
        where: { id: user.id },
        data: { name: "Updated" },
      });

      await prisma.$transaction(
        async (tx) => {
          const shouldBeUpdated = await tx.user.findUnique({
            where: { id: user.id },
          });
          console.log("shouldBeUpdated", shouldBeUpdated);
        },
        { isolationLevel: "ReadUncommitted" }
      );
    },
    { isolationLevel: "ReadUncommitted" }
  );
```

jonathanblade: Hmm... It's very interesting: 

> PostgreSQL's **Read Uncommitted** mode behaves like **Read Committed**. This is because it is the only sensible way to map the standard isolation levels to PostgreSQL's multiversion concurrency control architecture.

Also i found:

> The SQL standard to indicate that dirty read/non-repeatable read/phantom read are **possible** in certain transaction isolation level, but doesn't say they are **required**.

I'll rethink and rewrite the isolation level tests and get back.

RobertCraigie: one way to write the tests could just be to check what the isolation level of the transaction is, if that's even possible. e.g. `tx.query_raw('GET isolation_level')` (made that query up)

databases/tests/test_transactions.py
```diff
@@ -212,3 +212,61 @@ async def test_transaction_already_closed(client: Prisma) -> None:
...
+
+    assert len(results) == 1
+
+    row = results[0]
+    assert any(row)
+
+    level = next(iter(row.values()))
+    # The result can depends on the database, so we do upper() and replace()
+    level = str(level).upper().replace(' ', '_').replace('-', '_')
+    assert level == expected_level
```
jonathanblade, line 272: @RobertCraigie I updated the tests and found a bug in Prisma 😅 

P.S. Look please at [mypy (3.11)](https://github.com/RobertCraigie/prisma-client-py/actions/runs/7723046190/job/21052347877?pr=878) step why it doesn't pass.

RobertCraigie: Oh wow, thanks for reporting that!

The mypy failure is unrelated to your PR, I'll look into fixing it.

RobertCraigie: If you rebase against main, the mypy job should pass now.

jonathanblade: I'll update docs, then wait for a response from the Prisma team, and if all is ok, the PR will be ready to final review 🎉

jonathanblade: > I'll update docs, then wait for a response from the Prisma team, and if bug will be confirmed, then PR will be ready to final review 🎉

Comment: https://github.com/prisma/prisma/issues/22890#issuecomment-1923606830

RobertCraigie: of course mysql doesn't even have a way to query that 🤦 

I think it would be fine to just skip the tests for mysql if you haven't already.

jonathanblade: Sorry, I didn't know that...

It marked with [skipif](https://github.com/RobertCraigie/prisma-client-py/blob/577d1f112a83f909de0ec351e00675c9fd6fe0e1/databases/tests/test_transactions.py#L255-L264).

databases/utils.py
```diff
@@ -245,5 +254,8 @@ class RawQueries(BaseModel):
             SET title = 'updated title'
             WHERE id = 'sdldsd'
         """,
+        select_tx_isolation="""
+            PRAGMA read_uncommitted = 1
```
RobertCraigie, line 258: question: this query looks like it would mutate? Am I misunderstanding what this is doing?

docs/reference/transactions.md
```diff
@@ -114,6 +114,17 @@ In the case that this example runs successfully, then both database writes are c
...
 
+## Isolation levels
+
+By default, Prisma sets the isolation level to the value currently configured in the database. You can modify this
+default with the `isolation_level` argument (see [supported isolation levels](https://www.prisma.io/docs/orm/prisma-client/queries/transactions#supported-isolation-levels)).
+
+```py
+prisma.tx(
+    isolation_level=prisma.TransactionIsolationLevel.READ_UNCOMMITTED,
+)
```
RobertCraigie, line 125: nit: I think it would be helpful to include imports for this example & I think it's slightly incorrect as `prisma` can't be both the client variable and the module
```suggestion
from prisma import TransactionIsolationLevel

client.tx(
    isolation_level=TransactionIsolationLevel.READ_UNCOMMITTED,
)
```
The variable shadowing issue is also why I tend to try and stick to either using `db` or `client` as the `Prisma` instance, or to include a more complete example

```suggestion
from prisma import Prisma, TransactionIsolationLevel

client = Prisma()
client.tx(
    isolation_level=TransactionIsolationLevel.READ_UNCOMMITTED,
)
```
Up to you which you think is better :)

src/prisma/_transactions.py
```diff
@@ -3,23 +3,39 @@
...
+    'SyncTransactionManager',
+)
+
 
 _SyncPrismaT = TypeVar('_SyncPrismaT', bound='SyncBasePrisma')
 _AsyncPrismaT = TypeVar('_AsyncPrismaT', bound='AsyncBasePrisma')
 
 
+# See here: https://www.prisma.io/docs/orm/prisma-client/queries/transactions#supported-isolation-levels
+class TransactionIsolationLevel(StrEnum):
```
RobertCraigie, line 31: I'm somewhat torn between:
- Just defining this enum like you have with all options present
- Auto generating the enum to only include options that are supported by the current database

In general I've been leaning towards limiting code generation as much as possible but it might be worth it here? Do you have any opinions?

databases/sync_tests/test_transactions.py
```diff
@@ -201,3 +201,68 @@ def test_transaction_already_closed(client: Prisma) -> None:
...
+
+    assert len(results) == 1
+
+    row = results[0]
+    assert any(row)
+
+    level = next(iter(row.values()))
+    # The result can depends on the database, so we do upper() and replace()
+    level = str(level).upper().replace(' ', '_').replace('-', '_')
+    assert level == expected_level
```
RobertCraigie, line 268: minor nit (can leave as is): I think it would make sense to instead of defining `expected_level` as a single string, it is a dictionary that contains that's a map of database -> expected, e.g.

```py
pytest.param(
    prisma.TransactionIsolationLevel.SERIALIZABLE,
    {'postgresql': 'SERIALIZABLE', **etc},
    id='serializable',
    marks=pytest.mark.skipif(
        CURRENT_DATABASE == 'sqlite', reason='PRAGMA has only effect in shared-cache mode'
    ),
),
# ...
assert level == expected_levels[CURRENT_DATABASE]
```

RobertCraigie (commented):
Looking great! Just minor notes and a question about structuring the enum type.

databases/utils.py
jonathanblade: As far as I understand SQLite not support commands like `PRAGMA read_committed = 1` or `PRAGMA serializable = 1`. Only `PRAGMA read_uncommitted = 0` or `PRAGMA read_uncommitted = 1`.

Anyway execution of `PRAGMA` is [skipped](https://github.com/RobertCraigie/prisma-client-py/blob/110645e05695d62071245e4636d861bcaa15877a/databases/tests/test_transactions.py#L250) in tests.

src/prisma/_transactions.py
jonathanblade: If a user tries to start a transaction with an invalid/unsupported isolation level, they will receive a `500 Internal Server Error` from the Prisma, but the engine handle this case and raises [InvalidIsolationLevel](https://github.com/prisma/prisma-engines/blob/5.8.0/quaint/src/error.rs#L268-L269) error. I'm not sure, but it seems that `prisma-ts-client` should generate the enum.

databases/utils.py
jonathanblade: Forget the previous comment 🤦. SQLite doesn't have the way to query the current transaction isolation level.

I force pushed new version.

src/prisma/_transactions.py
jonathanblade: @RobertCraigie Finally, I chose this option:

> Auto generating the enum to only include options that are supported by the current database

PR updated.

PR #881, opened 2024-01-24
fix(types): remove definition of StrEnum as Enum

fisher60:
This addresses an issue with typing where string enums generated by prisma are not compatible with str typing. This specifically fixes an issue where Pyright checks would fail since prisma StrEnum has type Enum, which is not valid as a str.

## Change Summary

## Checklist

## Agreement

### DIFF
--- a/pipelines/requirements/dev.txt
+++ b/pipelines/requirements/dev.txt
@@ -9,3 +9,4 @@ typer==0.9.0
 rtoml==0.9.0
 GitPython
 distro
+StrEnum
--- a/pipelines/requirements/dev.txt
+++ b/pipelines/requirements/dev.txt
@@ -9,4 +9,3 @@ typer==0.9.0
 rtoml==0.9.0
 GitPython
 distro
-StrEnum
--- a/src/prisma/_compat.py
+++ b/src/prisma/_compat.py
@@ -338,16 +338,10 @@ else:
     except ImportError:
         nodejs = None
 
-
-if TYPE_CHECKING:
-    from enum import Enum
-
-    StrEnum = Enum
-else:
-    try:
-        from enum import StrEnum as StrEnum
-    except ImportError:
-        from strenum import StrEnum as StrEnum
+try:
+    from enum import StrEnum as StrEnum
+except ImportError:
+    from strenum import StrEnum as StrEnum
 
 
 def removeprefix(string: str, prefix: str) -> str:
--- a/src/prisma/_compat.py
+++ b/src/prisma/_compat.py
@@ -338,10 +338,17 @@ else:
     except ImportError:
         nodejs = None
 
-try:
-    from enum import StrEnum as StrEnum
-except ImportError:
-    from strenum import StrEnum as StrEnum
+
+if TYPE_CHECKING:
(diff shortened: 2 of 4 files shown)


RobertCraigie:
Thanks, can you add type checking tests for this? https://prisma-client-py.readthedocs.io/en/stable/contributing/contributing/#type-tests

RobertCraigie:
I think we still need the `TYPE_CHECKING` block, linters unfortunately do not like `try except` for imports like this :(

fisher60:
> Thanks, can you add type checking tests for this?

I am looking into this. Are there specific tests you have in mind that are not covered by other locations that utilize enums? i.e [lists.py](https://github.com/RobertCraigie/prisma-client-py/blob/main/typesafety/pyright/types/lists.py#L80) has several locations where it utilizes enums. Do we want more involved tests than are covered here? Or some similar tests explicitly made for enums?

fisher60:
Is there a benefit to maintaining the original try-except if the if-else statement handles checking python versions and importing the StrEnum?

RobertCraigie:
> Is there a benefit to maintaining the original try-except if the if-else statement handles checking python versions and importing the StrEnum?

That's a good point, I have a personal bias towards keeping the try-except but not for any logical reasons so feel free to just use the if statement 😅

RobertCraigie:
> I am looking into this. Are there specific tests you have in mind that are not covered by other locations that utilize enums? i.e [lists.py](https://github.com/RobertCraigie/prisma-client-py/blob/main/typesafety/pyright/types/lists.py?rgh-link-date=2024-01-25T00%3A05%3A15Z#L80) has several locations where it utilizes enums. Do we want more involved tests than are covered here? Or some similar tests explicitly made for enums?

I was thinking similar tests explicitly for enums, for example, the most important one would be a test case for the example code you shared before that broke when we were using `StrEnum = Enum`.

RobertCraigie:
Looks like mypy is reporting errors from the `strenum` library??

RobertCraigie:
The `pyright` lint failure in CI is likely because we're using 3.11 to run & install dependencies but targeting 3.7, IIRC we can't use 3.7 to install dependencies because some dev dependencies have dropped support for 3.7 and I didn't want to drop 3.7 support until there was a user-facing need to do so.

I think this'll be a problem for me as well locally as I use 3.11 when developing, so I think the solution would be to add `strenum` to our internal dev dependencies at `pipelines/requirements/dev.txt`.

RobertCraigie:
@fisher60 the lint error is because you need to `import StrEnum as StrEnum` 🙃

RobertCraigie:
hmm looks like mypy really doesn't like our usage of `strenum`... maybe we should just revert to the previous enum definition at the type-level as the only issues that had were at runtime, e.g.

```py
if TYPE_CHECKING:
  if sys.version >= (3, 11):
    from enum import StrEnum as StrEnum
  else:
    class StrEnum(str, Enum):
      ...
else:
  if sys.version >= (3, 11):
    from enum import StrEnum as StrEnum
  else:
    from strenum import StrEnum as StrEnum
```

RobertCraigie (approved):
Thank you!!

Issue #982, opened 2024-07-24
Support adding code to generated classes - e.g.: `model_config  = ConfigDict(protected_namespaces=())`


msabramo:
## Problem

https://github.com/BerriAI/litellm uses prisma-client-py and it happens to define some fields that start with `model_` and this causes Pydantic to complain with a bunch of warnings like this:

```
UserWarning: Field "model_max_budget" in LiteLLM_BudgetTable has conflict with protected namespace "model_".
```

## Suggested solution

It would be nice to be able to get rid of these warnings. To do this the generated code would have to have:
```python
model_config['protected_namespaces'] = ()
```

or something similar.

Ideally, this would be configurable so that most projects don't have to turn off this checking.

One possible way to solve this  is if prisma-client-py had a config option like `pyantic_disable_protected_namespaces` that would cause it to add the necessary code.

## Alternatives

Another option is a more general feature that allows adding any arbitrary Python code to the generated class. This would be more powerful but probably much harder to implement.

PR #983, opened 2024-07-24
Add `pydantic_disable_protected_namespaces` config setting

msabramo:
Fixes: GH-982

## Change Summary

Adds a `pydantic_disable_protected_namespaces` config setting. When this is true, it  adds:
```python
protected_namespaces=()
```

to the Pydantic `model_config`.

## Checklist

## Agreement

### DIFF
--- a/src/prisma/generator/models.py
+++ b/src/prisma/generator/models.py
@@ -506,6 +506,8 @@ class Config(BaseSettings):
     # https://github.com/prisma/prisma/issues/12442
     enable_experimental_decimal: bool = FieldInfo(default=False, env='PRISMA_PY_CONFIG_ENABLE_EXPERIMENTAL_DECIMAL')
 
+    pydantic_disable_protected_namespaces: bool = False
+
     # this seems to be the only good method for setting the contextvar as
     # we don't control the actual construction of the object like we do for
     # the Data model.
--- a/src/prisma/generator/templates/bases.py.jinja
+++ b/src/prisma/generator/templates/bases.py.jinja
@@ -21,6 +21,9 @@ class _PrismaModel(BaseModel):
             use_enum_values=True,
             arbitrary_types_allowed=True,
             populate_by_name=True,
+{% if generator.config.pydantic_disable_protected_namespaces %}
+            protected_namespaces=(),
+{% endif %}
         )
     elif not TYPE_CHECKING:
         from ._compat import BaseConfig


RobertCraigie:
Thanks for the PR.

Out of curiosity, do you know if renaming those fields would be difficult / not ideal?

msabramo:
> Thanks for the PR.
> 
> Out of curiosity, do you know if renaming those fields would be difficult / not ideal?

Well I'm not the maintainer of that project, but I'd say it would be difficult because we have a deployment of that software and we have a database with tables with those fields. That said I guess Prisma has migrations and maybe it would take of the renaming, but I don't have a lot of experience with it.

The problem is that `model` has 2 meanings here. Model in the Pydantic sense but also in LiteLLM, the model is referring to the model in Large Language Model (LLM). So `model` actually a pretty reasonable prefix to have -- it's just unfortunate that it overlaps with pydantic terminology.

RobertCraigie:
Thanks for the context, I will note that [it is possible to rename fields at the client level](https://www.prisma.io/docs/orm/prisma-schema/data-model/database-mapping#map-field--column-names) without causing db migrations :)

Granted that in this case, `model_` would really be the ideal field name anyway, I think this is a reasonable change.

I do wonder if this kind of setting should live on the model instead of the entire generator?
```prisma
@Python(pydantic_disable_protected_namespaces: true)
model LiteLLM_ProxyModelTable {
  model_id String @id @default(uuid())
  model_name String 
  litellm_params Json
  model_info Json? 
  created_at    DateTime               @default(now()) @map("created_at")
  created_by String
  updated_at    DateTime               @default(now()) @updatedAt @map("updated_at")
  updated_by String
}
```
but I don't feel strongly.

msabramo:
Yeah per-model seems nice. I think that would work.

msabramo:
How would I change this to be per-model?

RobertCraigie:
We have a [Lark](https://lark-parser.readthedocs.io/en/stable/how_to_use.html) grammar for parsing `@Python()` declarations.

So you'd have to update the `grammar/schema.lark` file to support booleans, run `nox -s lark`, then it should be a case of updating [this](https://github.com/RobertCraigie/prisma-client-py/blob/0cafaa78d7787f6bfc4ba06bc7ea40e5116f1c47/src/prisma/generator/models.py#L681) model type & using that config during codegen.

---

I just noticed my original example wouldn't be possible unfortunately, it'd have to be this fwiw
```prisma
/// @Python(pydantic_disable_protected_namespaces: true)
model LiteLLM_ProxyModelTable {
  model_id String @id @default(uuid())
  model_name String 
  litellm_params Json
  model_info Json? 
  created_at    DateTime               @default(now()) @map("created_at")
  created_by String
  updated_at    DateTime               @default(now()) @updatedAt @map("updated_at")
  updated_by String
}
```
---
additionally, I appreciate this is more work than you originally signed up for with this PR so I'm happy to take this over the line myself but unfortunately it won't land particularly fast.

msabramo:
Yeah, you can go ahead and take this over. I'm in no rush for this. Thanks!

Issue #998, opened 2024-08-08
Deserialization does not respect optional fields for special types for raw queries


AstreaTSS:
## Bug description

If a field can be null/`None` and is specially handled with the `DESERIALIZERS` variable found in `prisma/_raw_query` (ie BigInt, Decimal, Json), then any raw queries which has said field and is null will error out.

This only happens on 0.14.0 and above, as raw queries were reworked for that version.

## How to reproduce

1. Define a schema with an optional BigInt/Decimal/Json value:
```prisma
model PrismaBugExample {
  id          Int      @id @default(autoincrement())
  bad_var     BigInt?
}
```
2. Set up the tables and your Python program, and create an entry with a null/`None` value for `bad_var`:
```python
await PrismaBugExample.prisma().create(data={"bad_var": None})
```
3. Make a raw query getting that entry:
```python
data = await PrismaBugExample.prisma().query_first("SELECT * FROM PrismaBugExample")
```
The bug should occur at this step.

## Expected behavior

For raw queries including these nullable types to work.

## Prisma information

Schema: https://github.com/AstreaTSS/PYTHIA/blob/8873fd94c770c9faa503c70201c01af3a0390943/schema.prisma
Models (there is some post-generation adjustments that I do to them - it has no affect on this bug but I might as well include everything): https://github.com/AstreaTSS/PYTHIA/blob/8873fd94c770c9faa503c70201c01af3a0390943/common/models.py
Query:
- Actual Query: https://github.com/AstreaTSS/PYTHIA/blob/8873fd94c770c9faa503c70201c01af3a0390943/common/models.py#L434-L452 
- Execution: https://github.com/AstreaTSS/PYTHIA/blob/8873fd94c770c9faa503c70201c01af3a0390943/common/models.py#L122-L124

```prisma

```

## Environment & setup

- OS: Windws
- Database: PostgreSQL
- Python version: 3.12.3
- Prisma version:

```
prisma                  : 5.17.0
prisma client python    : 0.14.0
platform                : debian-openssl-1.1.x
expected engine version : 393aa359c9ad4a4bb28630fb5613f9c281cde053
installed extras        : []
install path            : /home/astrea/Documents/PYTHIA/env/lib/python3.12/site-packages/prisma
binary cache dir        : /home/astrea/.cache/prisma-python/binaries/5.17.0/393aa359c9ad4a4bb28630fb5613f9c281cde053
```

## Internal cause

[...]

Issue #714, opened 2023-03-02
Add support for `Prisma.DbNull` equivalent


RobertCraigie:
## Problem

https://discord.com/channels/933860922039099444/1080804462970535996/1080804462970535996

## Suggested solution

Implement equivalent of https://www.prisma.io/docs/concepts/components/prisma-client/working-with-fields/working-with-json-fields#using-null-values

ezorita:
@RobertCraigie I'd support the development of this feature, since we found ourselves with the issue of not being able to NULL optional Json fields.

PR #1000, opened 2024-08-08
fix(raw queries): correct None handling for all fields

AstreaTSS:
## Change Summary

See https://github.com/RobertCraigie/prisma-client-py/issues/998 for the most part.

This PR goes in and makes the check mentioned in the above issue happen earlier.

## Checklist

## Note
Adding tests for JSON was not done due to technical difficulties I had when testing. Apparently, optional JSON fields can't accept a `None` value when being created, according to Prisma, despite the fact that I see it as [optional in the schema used](https://github.com/RobertCraigie/prisma-client-py/blob/f1ab3b183cbfaf2d0c7c0c9bce0aeec346236841/databases/templates/schema.prisma.jinja2#L87). I'll leave what errors I got below, though it may be just a me thing.

```python
record = await client.types.create({'json_obj': None})
```

```
prisma.errors.MissingRequiredValueError: Unable to match input value to any allowed input type for the field. Parse errors: [Unable to match input value to any allowed input type for the field. Parse errors: [`data.json_obj`: A value is required but not set, `data.json_obj`: A value is required but not set], Unable to match input value to any allowed input type for the field. Parse errors: [`data.json_obj`: A value is required but not set, `data.json_obj`: A value is required but not set]]
```

## Agreement

### DIFF
--- a/databases/sync_tests/types/raw_queries/test_bigint.py
+++ b/databases/sync_tests/types/raw_queries/test_bigint.py
@@ -9,14 +9,17 @@ from ...._compat import LiteralString
 
 class Queries(BaseModel):
     select: LiteralString
+    select_null: LiteralString
 
 
 _mysql_queries = Queries(
     select='SELECT * FROM Types WHERE `bigint` = ?',
+    select_null='SELECT * FROM Types WHERE `optional_bigint` IS NULL',
 )
 
 _postgresql_queries = Queries(
     select='SELECT * FROM "Types" WHERE bigint = $1',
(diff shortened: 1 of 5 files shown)


RobertCraigie:
> Adding tests for JSON was not done due to technical difficulties I had when testing

ah I think this is because of #714

RobertCraigie (approved):
Thanks so much for the PR! Sorry for the regression in the last release.

I pushed a commit adding tests for the synchronous client

PR #1002, opened 2024-08-09
Explicitly use utf-8 in path.read_text

AmaseCocoa:
## Change Summary
Changed `path.read_text` to explicitly use utf-8 to avoid `UnicodeEncodeError` that may be caused by cp932 in Japanese Windows environments.

Only src/prisma/_config.py was changed.

(This is my first time contributing to another repository, so there may be some clumsiness, but I'd appreciate it if you could let me know if that's the case.)
## Checklist

## Agreement

### DIFF
--- a/src/prisma/_config.py
+++ b/src/prisma/_config.py
@@ -108,7 +108,7 @@ class Config(DefaultConfig):
             path = Path('pyproject.toml')
 
         if path.exists():
-            config = tomlkit.loads(path.read_text()).get('tool', {}).get('prisma', {})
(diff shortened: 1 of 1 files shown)


RobertCraigie (commented):
Thanks for the PR!

It'd be great if you could add a unit test for this case to `tests/test_config.py`. We have some existing tests for config loading there which you can take inspiration from :)

AmaseCocoa:
I added Unit Test (`test_with_file`) to `tests/test_config.py` :)

Issue #52, opened 2021-08-26
Add support for full text search for PostgreSQL


RobertCraigie:
[https://www.prisma.io/docs/concepts/components/prisma-client/full-text-search](https://www.prisma.io/docs/concepts/components/prisma-client/full-text-search)

## Suggested solution

Something like the following, `search` should be added to `StringFilter` and must be an instance of `String`.

Should be noted that I'm not stuck on `String` being the name.

```py
# NOTE: actual API is still TODO
from prisma.querying import String

await client.post.find_first(
    where={
        'content': {
            'search': String.contains('cat', 'dog'),
        },
    },
)
await client.post.find_first(
    where={
        'content': {
            # for anything we don't explicitly support
            'search': String.raw('fox \| cat'),
        },
    },
)
```

kasem-sm:
Any updates on this?

ntindle:
subscribing here :)

AdeelK93:
the `String` idea is interesting, but may prove to be challenging to robustly implement, and keep consistent with the TS client.
alternatively, could this be achieved via adding `search` to `StringFilter`?

RobertCraigie:
Yeah `String` could definitely be too involved.

Just supporting a `str` like the TS client does sounds good to me for now. We could always add a more expressive API later if needed.

RobertCraigie:
Implemented in #1003

PR #1003, opened 2024-08-10
feat(client): Add support for full text search

AdeelK93:
## Change Summary

* #52
* Supports full text search in where clause for postgresql and mysql
* Supports relevance ordering - which requires tracking `active_provider` in `Model` context

## Checklist

## Agreement

### DIFF
--- a/docs/reference/operations.md
+++ b/docs/reference/operations.md
@@ -332,6 +332,8 @@ post = await db.post.find_first(
 )
 ```
 
+Please visit Prisma's [documentation](https://www.prisma.io/docs/orm/prisma-client/queries/full-text-search) for more information on full-text search.
+
 #### Integer Fields
 
 ```py
--- a/docs/reference/operations.md
+++ b/docs/reference/operations.md
@@ -301,6 +301,7 @@ post = await db.post.find_first(
 
 !!! warning
     Case insensitive filtering is only available on PostgreSQL and MongoDB
+    Full-text search is only available on PostgreSQL and MySQL
 
 ```py
 post = await db.post.find_first(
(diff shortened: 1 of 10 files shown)


src/prisma/generator/schema.py
```diff
@@ -80,7 +80,8 @@ class Schema(BaseModel):
 
     @classmethod
     def from_data(cls, data: AnyData) -> 'Schema':
-        models = [Model(info=model) for model in data.dmmf.datamodel.models]
+        active_provider = data.datasources[0].active_provider
+        models = [Model(info=model, active_provider=active_provider) for model in data.dmmf.datamodel.models]
```
RobertCraigie, line 84: nit: I think plumbing the provider through like this is a little awkward, we have a similar problem in other places and the solution I came up with is to set a [`ContextVar`](https://github.com/RobertCraigie/prisma-client-py/blob/64e0ddc6f137c57960930446f727160bc1cf3788/src/prisma/generator/models.py#L123). I don't think that solution is *amazing* either but I think it is better than having to pass through information like this.
```py
from .models import data_ctx

# ...

if data_ctx.get().datasources[0].active_provider: ...
```

docs/reference/operations.md
```diff
@@ -319,6 +320,7 @@ post = await db.post.find_first(
             'endswith': 'must end with string',
             'in': ['find_string_1', 'find_string_2'],
             'mode': 'insensitive',
+            'search': 'full-text query',
```
RobertCraigie, line 323: nit: might be nice to include an example of the special full-text syntax & link to either Prisma or psql / mysql docs for the syntax.

RobertCraigie (approved):
Thanks so much!

RobertCraigie:
Thanks so much for all your contributions Adeel!

Issue #1008, opened 2024-08-11
Improve error message when running prisma with an outdated client


AdeelK93:
## Problem

When upgrading to a newer version of prisma python client - let's say, 0.13.1 to 0.14.0, if you forget to run `prisma generate` again, you'll get an error message like this

>   File "/Users/adeel/Documents/GitHub/leverage-optimizer/backend/utils/db.py", line 36, in get_db
    _db = Prisma(auto_register=True, datasource={"url": get_db_uri(application_name)})
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/adeel/Library/Caches/pypoetry/virtualenvs/leverage-optimizer-cJkWU15t-py3.12/lib/python3.12/site-packages/prisma/client.py", line 156, in __init__
    self._set_generated_properties(
TypeError: BasePrisma._set_generated_properties() missing 1 required keyword-only argument: 'preview_features'

It's not very obvious based on this `preview_features` message that this error actually stems from an outdated client.

## Suggested solution

The prisma library already knows `prisma.__version__`. We should also track the version used for generating the client - and if it's missing or mismatched, raise an error.

PR #1018, opened 2024-08-18
feat(client): better version mismatch handling

RobertCraigie:


jonathanblade:
I just mention issue https://github.com/RobertCraigie/prisma-client-py/issues/1008 for new developers so they can see that this feature is already in development 😃

PR #1027, opened 2024-09-06
docs: update the link in advanced.md

AuroraTea:
## Change Summary

There is a link to the Prisma Client Go documentation, but it no longer works because the path to the other document has changed.

## Agreement


RobertCraigie (approved):
Thanks!

PR #1035, opened 2024-09-28
feat(engine): json protocol

### LINKED ISSUE #707: Refactor internal query builder to use the JSON protocol
Prisma are working on removing the internal GraphQL protocol and moving to a JSON-based approach instead which will simplify a lot of internal code and potentially result in the ability to completely remove certain codegen-metadata resulting in smaller package sizes in applications where that matters.

This would also remove the requirement to wrap json values in `Json(...)`!

ProgramRipper:
closed #707, #748

## Change Summary

A naive implementation of JSON protocol based on #748.

## Checklist

## Agreement


databases/tests/test_errors.py
```diff
@@ -39,19 +39,19 @@ async def test_field_not_found_error(client: Prisma) -> None:
...
-    with pytest.raises(
-        FieldNotFoundError,
-        match=r'Field \'foo_field\' not found in enclosing type \'Post\'',
-    ):
-        await CustomPost.prisma().find_first()
+#     with pytest.raises(
+#         FieldNotFoundError,
+#         match=r'Field \'foo_field\' not found in enclosing type \'Post\'',
+#     ):
+#         await CustomPost.prisma().find_first()
```
ProgramRipper, line 54: > [!CAUTION]
> `test_field_not_found_error_selection` skipped

JSON protocol does not explicitly select all fields, so it won't raise a `FieldNotFoundError` here.

However we could:
- explicitly select all fields (cons: this could be hard to impl and bring overhead to serializer, prisma-engine and maybe even db); or...
- leave it to pydantic as it will raise validation errors if data doesn't match the model.

Issue #314, opened 2022-03-03
Add support for composite types


RobertCraigie:
## Problem

In 3.10 Prisma added support for embedded types, this looks like this in the schema

```prisma
model User {
  id Int @id @default(autoincrement())
  name String
  photo Photo
}

type Photo {
  width  Int
  height Int
  data   Bytes
}
```

This is currently only supported for MongoDB although there are plans to add support for this to other database providers in the future.

## Suggested solution

I have not looked into the API that Prisma provides for this yet but I imagine that this will be very similar to relational fields.

https://www.prisma.io/docs/concepts/components/prisma-client/composite-types

farzad-salimijazi:
@RobertCraigie, Is this supported using the Prisma python client yet?

RobertCraigie:
@farzad-salimijazi Unfortunately it's not supported yet. This is a more difficult feature to implement as we currently only run tests against SQLite and PostgreSQL, so before I can even start working on this I need to figure out a good solution for testing MongoDB as well.

It's also difficult because this is an entirely new concept which we'll need to figure out how to design internally.

In the meantime you can either fall back to using a `Json` field or define it as a model instead, although that could be a bit annoying to work with.

timreibe:
+1

arthur-fontaine:
+1

arthur-fontaine:
I have just opened a PR https://github.com/RobertCraigie/prisma-client-py/pull/1045 to add support for composite types. The PR is not finished yet, at the moment it only generates client types. I need to write e2e tests to know what else needs to be done.

PR #1045, opened 2024-11-17
feat(generator): initial support for typing of Composite Types

arthur-fontaine:
## Change Summary

This is a draft PR that aims to bring support for [composite types](https://www.prisma.io/docs/orm/prisma-client/special-fields-and-types/composite-types) (#314).

As I don't know the project well, I am open to receive reviews and guidance on this.

## Checklist

## Agreement

### DIFF
--- a/src/prisma/generator/models.py
+++ b/src/prisma/generator/models.py
@@ -2,6 +2,7 @@ import os
 import sys
 import enum
 import textwrap
+import warnings
 import importlib
 from typing import (
     TYPE_CHECKING,
@@ -652,19 +653,13 @@ class DMMF(BaseModel):
 class Datamodel(BaseModel):
     enums: List['Enum']
     models: List['Model']
+    types: List['Types']
 
-    # not implemented yet
-    types: List[object]
 
-    @field_validator('types')
-    @classmethod
-    def no_composite_types_validator(cls, types: List[object]) -> object:
-        if types:
-            raise ValueError(
-                'Composite types are not supported yet. Please indicate you need this here: https://github.com/RobertCraigie/prisma-client-py/issues/314'
-            )
-
-        return types
+class Types(BaseModel):
+    name: str
+    db_name: Optional[str] = FieldInfo(alias='dbName')
+    fields: List['TypeField']
 
 
 class Enum(BaseModel):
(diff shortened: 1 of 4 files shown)


arthur-fontaine:
@RobertCraigie Can you clarify the purposes of the models defined in `types.py` and those defined in `models.py`?
After a little more digging, I think that the composite types I am injecting in `types.py` [a7f6263](https://github.com/RobertCraigie/prisma-client-py/pull/1045/commits/a7f62630a362d1e67f9d8736b86d7215dea0da78) should be injected in their own file `composite_types.py`.

RobertCraigie:
thanks for working on this!!

> Can you clarify the purposes of the models defined in types.py and those defined in models.py?
After a little more digging, I have the feeling that the composite types I am injecting in types.py [a7f6263](https://github.com/RobertCraigie/prisma-client-py/pull/1045/commits/a7f62630a362d1e67f9d8736b86d7215dea0da78) should be injected in their own file composite_types.py.

Ah yeah `models.py` is just for the models defined in the schema and `types.py` is for every other auxilary type we need to generate, e.g. the update params types.

Other types that come directly from the schema have their own file as well, e.g. `enums.py` so outputting to a separate `prisma/composite_types.py` file sounds good to me!

arthur-fontaine:
In fact, I don't know if I really need to write tests specifically for MongoDB, since the query is handled by the Prisma engine after all. Maybe it is better to write tests for the query builder?
What do you think about this?

RobertCraigie:
yeah the tests you write for this should just treat composite types as a DB feature and then you'd write tests for the client query methods, e.g. see how the enum tests are written
- https://github.com/RobertCraigie/prisma-client-py/blob/main/databases/tests/test_enum.py
- https://github.com/RobertCraigie/prisma-client-py/blob/05808fc6fbe6fccd3465a92b8bae81af80caa8aa/databases/utils.py#L16
- https://github.com/RobertCraigie/prisma-client-py/blob/05808fc6fbe6fccd3465a92b8bae81af80caa8aa/databases/constants.py#L97

arthur-fontaine:
The problem is that there are too many differences between relational databases and MongoDB when using Prisma. This led me to create a new `tests_mongodb`.
However, since tests for operations like `create` and `find` are already implemented for relational databases—and both relational databases and MongoDB rely on the same Prisma Query Engine—I’m questioning whether separate tests for MongoDB are truly necessary.

Would it make more sense to focus on unit tests for composite types in `test_builder.py` instead?

PR #1053, opened 2024-12-13
Fix platform check

m3at:
`cat /etc/os-release` is not always available, the prisma client broke when trying to deploy in an Ubuntu 22.04 VM. This uses the standard library's `platform` instead.

## Change Summary

## Checklist

## Agreement

### DIFF
--- a/src/prisma/binaries/platform.py
+++ b/src/prisma/binaries/platform.py
@@ -32,15 +32,12 @@ def linux_distro() -> str:
 
 
 def _get_linux_distro_details() -> Tuple[str, str]:
-    process = subprocess.run(['cat', '/etc/os-release'], stdout=subprocess.PIPE, check=True)
-    output = str(process.stdout, sys.getdefaultencoding())
-
-    match = re.search(r'ID="?([^"\n]*)"?', output)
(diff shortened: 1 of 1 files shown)


src/prisma/binaries/platform.py
```diff
@@ -32,15 +30,12 @@ def linux_distro() -> str:
...
-
-    match = re.search(r'ID="?([^"\n]*)"?', output)
-    distro_id = match.group(1) if match else ''
-
-    match = re.search(r'ID_LIKE="?([^"\n]*)"?', output)
-    distro_id_like = match.group(1) if match else ''
-    return distro_id, distro_id_like
+    if hasattr(_platform, 'freedesktop_os_release'):
+        # For python >= 3.10
+        distro = _platform.freedesktop_os_release()
```
dmechea, line 35: is this because of prisma? what are you checking? care to explain like im 5? :smile:

m3at: The same check as was done through reading `/etc/os-release`, but using the stdlib's `platform.freedesktop_os_release()`, though as it's py3.10 and above only (and this repo supports 3.8), I use the older `platform.linux_distribution()` for older pythons

